PR #6, opened 2026-03-30
feat: add upstream sync check hook

tarilabs:
demo:

Screenshot 2026-03-30 at 21 35 32


jwforres:
@tarilabs now that we've made this part of the [skills registry](https://github.com/opendatahub-io/skills-registry?tab=readme-ov-file#quick-start) and its installable via the plugin mechanisms it should be updating through that. Can you try that out and see if this is still needed? That would be the preferred installation mechanism moving forward vs. directly git cloning this repo which should be more for people developing the skill.

tarilabs:
I concur as the skill marketplace was not available at the time, so now this is obsolete either way.

PR #67, opened 2026-04-07
Fix batch creation and auto-fix completion for cross-model reliability

astefanutti:
## Summary

- Use parallel Agents with pre-assigned IDs for Phase 1 batch creation (fixes Sonnet)
- Add `--rfe-id` flag to `/rfe.create` for deterministic ID assignment
- Add batch completion guardrails to `/rfe.auto-fix` (prevents early stopping)

## Problem

Three issues found via the [agent-eval-harness](https://github.com/opendatahub-io/agent-eval-harness) during cross-model benchmarking:

### 1. Phase 1: Sonnet stops after first entry

The original instruction ("invoke `/rfe.create` as an inline Skill") fails with Sonnet — it stops after processing the first entry because inline Skill calls don't maintain the loop across returns.

| Approach | Opus | Sonnet |
|----------|------|--------|
| Sequential inline Skill | Works (fragile) | **Fails** — creates 1-5 RFEs from 20, then exits |
| **Parallel Agents + pre-assigned IDs** | **All 20, deterministic** | **All 20, deterministic** |

### 2. Phase 1: Non-deterministic ID assignment

Parallel Agents race to `next_rfe_id.py`, so batch entry index doesn't match RFE ID. This breaks eval harness case-to-output mapping, baseline diffing, and traceability.

**Fix**: Pre-allocate IDs with `next_rfe_id.py N`, pass `--rfe-id RFE-NNN` to each Agent.

### 3. Phase 2: Auto-fix stops after first batch

Without explicit guardrails, the LLM sometimes stops after processing batch 1 of 4, outputting a summary instead of continuing. Observed in both Opus and Sonnet during long runs with context compression.

**Fix**: Add CRITICAL instruction requiring all batches to be processed, plus a continuation check between batches that re-reads state from disk.

### rfe.speedrun (Phase 1)
- Count entries with a script (not LLM eyeballing — it miscounts ~10% of the time)
- Pre-allocate all IDs via `next_rfe_id.py N`
- Launch parallel Agents with `--rfe-id RFE-NNN` per entry
- Add guardrail: never delete/re-create task files during Phase 1

### rfe.create
- Accept `--rfe-id <ID>` flag to use a pre-assigned ID instead of calling `next_rfe_id.py`
- Backward compatible: omitting `--rfe-id` preserves original behavior

### rfe.auto-fix
- Add CRITICAL instruction: must process every batch, no exceptions
- Add continuation check at batch boundary: re-read state from disk, verify `current_batch < total_batches`, loop back if more remain

## Test plan

- [x] Opus with 20 entries — all 20 RFEs created, deterministic IDs
- [x] Sonnet with 20 entries — all 20 RFEs created
- [x] Auto-fix processes all 4 batches without stopping early
- [ ] Verify `--rfe-id` doesn't break interactive `/rfe.create` (no flag = original behavior)
- [ ] Verify single-idea mode (Mode C) still works without `--rfe-id`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.claude/skills/rfe.create/SKILL.md
+++ b/.claude/skills/rfe.create/SKILL.md
@@ -13,6 +13,7 @@ Parse `$ARGUMENTS` for:
 - `--headless`: Skip clarifying questions (Step 2) — generate RFEs directly from the input
 - `--priority <value>`: Override default priority (Blocker, Critical, Major, Normal, Minor)
 - `--labels <comma-separated>`: Labels to apply to created RFEs
+- `--rfe-id <ID>`: Pre-assigned RFE ID. When provided, use this ID instead of calling `next_rfe_id.py` in Step 4. The placeholder file already exists.
 - Remaining arguments: the problem statement / idea text
 
 If `--headless` is present, skip Step 2 entirely and proceed directly from Step 1 to Step 3 using the provided input.
@@ -62,9 +63,9 @@ Key rules:
 
 ## Step 4: Write Artifacts
 
-For each RFE, allocate an ID first, then write the markdown body and set frontmatter.
+For each RFE, determine its ID, then write the markdown body and set frontmatter.
 
-Allocate IDs atomically (prevents collisions with parallel runs):
+If `--rfe-id` was provided, use that ID (the placeholder file already exists). Otherwise, allocate IDs atomically:
 
 ```bash
 python3 scripts/next_rfe_id.py <count>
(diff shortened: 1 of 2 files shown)


jwforres:
@astefanutti it looks like this has conflicts with your other PR that I just merged #66

astefanutti:
@jwforres I'm iterating a bit on this to get it work also with Sonnet, and I'll rebase it once I'll have thoroughly tested.

astefanutti:
@jwforres it's updated now with what I had to do to make #74 and the `rfe.speedun` skill work with Sonnet.

I'm not sure if we want to go as far. Maybe I can wait with the latest dispatcher development that may improve how Sonnet perform.

PR #76, opened 2026-04-09
feat: init work for dedupe when creating a rfe

syntaxsdev:
### Description:  Add duplicate detection to /rfe.create                                                                          
                                                                                                                  
  When a user runs /rfe.create, the skill now checks for existing RFEs that may overlap with the new idea before  
  asking clarifying questions. This prevents duplicate work and surfaces related efforts early.
                                                                                                                  
  How it works:                                             
  - Extracts key phrases from the problem statement                                                               
  - Searches a local cache of open RFE summaries (refreshed every 4 hours on-demand based on TTL)
  - Falls back to JQL text search if cache results are sparse                                                     
  - Presents potential duplicates as a table with links      
  - User decides: connect to existing RFE or proceed with creation                                                
  - Non-blocking: if search fails or returns nothing, creation continues normally

### DIFF
--- a/.claude/agents/rfe-scorer.md
+++ b/.claude/agents/rfe-scorer.md
@@ -1,0 +1,8 @@
+---
+name: rfe-scorer
+description: Scores a single RFE issue against quality rubric. Restricted to Read and Write only to prevent prompt injection exfiltration.
+tools: Read, Write
+permissionMode: acceptEdits
+---
+
+You are an RFE quality assessor. You score Jira issues against a rubric. You can only read files and write results — you have no other capabilities.
--- a/.claude/skills/rfe.create/SKILL.md
+++ b/.claude/skills/rfe.create/SKILL.md
@@ -31,6 +31,36 @@ If `artifacts/rfe-rubric.md` exists (either already present or just exported), r
 
 If the rubric is still not available after the bootstrap attempt, proceed with the built-in question flow below.
 
+## Step 1.5: Check for Potential Duplicates
+
+Before asking clarifying questions, check whether similar RFEs already exist in Jira. This prevents duplicate work and helps the PM connect to existing efforts.
+
+1. Extract 3-5 key phrases from the problem statement that capture the core need. Focus on domain-specific terms (product names, capabilities, user actions) rather than generic words like "support", "enable", "improve".
+
+2. Run the duplicate search:
+
+```bash
+python3 scripts/dedup_search.py search "<problem statement text>" --keywords "phrase1,phrase2,phrase3" --max-results 10
+```
+
+3. Parse the JSON output. If matches were found:
+
+   **Interactive mode**: Present the top 5 matches as a table:
+
+   | Key | Summary | Link |
+   |-----|---------|------|
+   | RFE-1234 | Example summary | [View](https://...) |
+
+   Then ask the user:
+   > "I found existing RFEs that may overlap with your idea. Do any of these already cover what you need? (yes/no/not sure)"
+
+   - If **yes**: Ask which one(s). Suggest they comment on the existing RFE instead of creating a new one. Offer to continue creating a new RFE anyway if they want.
+   - If **no** or **not sure**: Continue to Step 2.
+
+   **Headless mode (`--headless`)**: Log the matches to stderr but do not block. Proceed directly to Step 3.
+
+4. If the search fails (error in JSON output) or returns no matches, proceed silently to Step 2. Never block RFE creation due to a search failure.
+
 ## Step 2: Clarifying Questions
 
 Before generating RFEs, ask the PM clarifying questions to fill gaps. Ask 2-5 questions maximum — only ask what you cannot reasonably infer from the input. Focus on:
--- a/.claude/skills/rfe.create/SKILL.md
+++ b/.claude/skills/rfe.create/SKILL.md
@@ -35,15 +35,41 @@ If the rubric is still not available after the bootstrap attempt, proceed with t
 
 Before asking clarifying questions, check whether similar RFEs already exist in Jira. This prevents duplicate work and helps the PM connect to existing efforts.
 
-1. Extract 3-5 key phrases from the problem statement that capture the core need. Focus on domain-specific terms (product names, capabilities, user actions) rather than generic words like "support", "enable", "improve".
+> **Design reference**: see `docs/duplicate-detection.md` for the full design (two-tier cache, opt-in build rationale, `source` output values, known gaps). The skill instructions below are the operational flow — the design doc explains *why*.
 
-2. Run the duplicate search:
+1. Extract 3-5 key phrases from the problem statement that capture the core need. Focus on domain-specific terms (product names, capabilities, user actions) rather than generic words like "support", "enable", "improve". (`--keywords` is optional — the script has a stopword-filtered fallback — but explicit phrases produce much better matches.)
 
-```bash
-python3 scripts/dedup_search.py search "<problem statement text>" --keywords "phrase1,phrase2,phrase3" --max-results 10
-```
+2. **Choose a search mode.** The duplicate search uses a local cache of open RFE summaries for speed. Building that cache from scratch paginates the full Jira project and can take 10+ seconds on large projects, so we let the user opt in:
 
-3. Parse the JSON output. If matches were found:
+   First, probe cache state:
+
+   ```bash
+   python3 scripts/dedup_search.py cache-info --json
+   ```
+
+   - **Headless mode (`--headless`)**: skip the probe. Always pass `--headless --recent-only` to the search in step 3 so CI never blocks on a full cache build.
+   - **Interactive mode, cache is "fresh"** (per the JSON `fresh: true` field): silently use it. Skip the opt-in prompt.
+   - **Interactive mode, cache is missing or stale**: ask the user via `AskUserQuestion`:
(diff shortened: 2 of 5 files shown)


jwforres:
### 1. JQL injection risk in `_search_jql` (scripts/dedup_search.py:164-166)

Keywords are user-provided (or LLM-extracted from user text). The escaping only handles double quotes:

```python
clean = kw.replace('"', '\\"')
escaped.append(f'"{clean}"')
```

JQL special characters like `\`, `{`, `}`, `[`, `]` are not escaped. A keyword containing a backslash could break the query or cause unexpected behavior. Consider sanitizing more aggressively or using Jira's REST API body-based JQL (POST to search endpoint) which avoids URL-encoding issues.

### 2. No test coverage for `search()` or `_search_jql`

The tests cover cache-based search and cache management well, but the main `search()` function and the JQL fallback path have no unit tests (even with mocked HTTP). These are the most complex code paths -- the merge/dedup logic and the tier-2 fallback trigger (`len(cache_results) < 3`) should be tested.

### 3. `_extract_fallback_keywords` is used but not referenced in the SKILL.md flow

The skill instructions always tell the LLM to pass `--keywords`. The fallback extractor in `cmd_search` kicks in when `--keywords` is omitted, but the skill never exercises that path. This is fine as defensive code, but worth noting -- if the intent is that the LLM always provides keywords, consider whether the fallback is dead code or whether the skill instructions should mention it as optional.

### 4. `require_env()` called inside `_cache_is_fresh` creates hidden side effects (scripts/dedup_search.py:80-86)

`_cache_is_fresh` looks like a pure check -- "give me a cache dict, I'll tell you if it's fresh" -- but it secretly depends on the process environment:

```python
def _cache_is_fresh(cache):
    age = _cache_age_hours(cache)
    if age is None or age > CACHE_TTL_HOURS:
        return False
    server, _, _ = require_env()       # <-- reads JIRA_SERVER from os.environ
    if server and cache.get("server") != server:
        return False
    return True
```

Problems with this:

- **Testing is fragile.** Every test that calls `_cache_is_fresh` must remember to set `JIRA_SERVER` via monkeypatch, or the server comparison silently does the wrong thing. The existing tests do this correctly, but it's easy to miss in future tests.
- **Reusability is limited.** You can't call `_cache_is_fresh(cache)` in a context where you already have the server string in hand -- it always re-reads from env.
- **Callers can't see the dependency.** The function signature says "give me a cache dict" but the answer also depends on `os.environ`. Nothing in the interface signals this.

Suggested fix -- pass server explicitly:

```python
def _cache_is_fresh(cache, server=None):
    age = _cache_age_hours(cache)
    if age is None or age > CACHE_TTL_HOURS:
        return False
    if server and cache.get("server") != server:
        return False
    return True
```

Then the caller (which already has the server from `require_env()`) passes it in. The function becomes a pure check with no hidden side effects.

### 5. Inconsistent output format between CLI subcommands

`search` returns structured JSON to stdout:

```json
{"source": "cache", "matches": [...], "error": null}
```

`cache-info` returns human-readable text to stdout:

```
Cache file: tmp/dedup-cache.yaml
Refreshed: 2026-04-09T14:00:00+00:00
Age: 1.5h (TTL: 4h, fresh)
Issues: 342
```

This means if someone wanted to script around the tool -- say, check if the cache is stale before running a batch of searches -- they'd have to parse free-form text instead of JSON. Adding a `--json` flag (or just making JSON the default like `search` does) would keep the CLI consistent and scriptable. Low priority.

## Architectural: Cache Performance at Scale

Other teams adopting this tool have projects with thousands of open issues. The current design — full cache rebuild on every TTL expiry, blocking the user while paginating — won't scale. The three recommendations below work together as a coherent set.

### A. Don't block on cache build — make it opt-in (interactive) or skipped (headless)

The current code silently blocks while paginating through all open issues on the first search. For a project with 5,000 issues at 100/page, that's ~50 sequential API calls — potentially a minute of dead time before the user sees anything.

**Interactive mode (no cache exists):** Prompt the user:

> I can build a local cache of open RFEs to check for duplicates. This takes a moment but improves future searches.
> 1. Yes, build cache (Recommended)
> 2. No, just check recent issues

- **Option 1:** Builds the full cache (same as today), then searches it. Future searches within the TTL are instant.
- **Option 2:** Skips cache entirely, runs a targeted JQL query scoped to recent issues — e.g., `project = RHAIRFE AND statusCategory != Done AND created >= -30d AND text ~ "keywords"`. Fast, but only catches recent duplicates.

**Interactive mode (cache exists, fresh):** No prompt — use the cache silently.

**Interactive mode (cache exists, stale):** Silently refresh. The user already opted in, and with incremental refresh (recommendation B) this becomes fast.

**Headless mode:** Never prompt, never block on a full build. Use the cache if fresh, otherwise fall straight to JQL. A CI run should never wait a minute to build a dedup cache.

This ensures the expensive operation is always opt-in, and the "no" path is still useful — just with a narrower search window.

### B. Incremental cache refresh instead of full replacement

The cache already stores `refreshed_at`. Use it to fetch only the delta:

```
project = RHAIRFE AND statusCategory != Done AND updated >= "2026-04-13 10:00"
```

First run is still a full fetch, but subsequent refreshes only pull issues changed since the last refresh — typically 1-2 pages instead of 50. This drops a minute-long refresh to seconds.

The complication is handling closures: issues that moved to Done since the last refresh need to be *removed* from the cache. Options:
- A second query: `project = RHAIRFE AND statusCategory = Done AND updated >= "<last_refresh>"` to find keys to evict
- Periodic full reconciliation (e.g., full refresh every 24h, incremental every 4h)

Either approach is a major improvement over always doing a full fetch.

### C. Switch the cache format from YAML to JSON

The cache is a flat dict of key-summary pairs — no anchors, no multi-line strings, nothing that needs YAML's features. `json.load`/`json.dump` is 10-20x faster than PyYAML for large dicts, and it's in the stdlib — eliminates the `try/except ImportError` guard for the `yaml` module. Drop-in change, no behavior difference.

## Minor Nits

- The `_extract_fallback_keywords` stopword list (~60 words) is recreated as a set on every call. Consider moving it to a module-level constant.

jwforres:
kw.lower() is called inside the inner loop.                                                                             
                  
  `hits = sum(1 for kw in keywords if kw.lower() in summary_lower)             `                                               
                                                                                                                             
  kw.lower() allocates a new string on every iteration of the outer loop. For 5,000 issues and 5 keywords, that's 25,000     
  string allocations for values that never change. Should be hoisted:                                                        
  ```                                                                                                                           
  keywords_lower = [kw.lower() for kw in keywords]  # once
  for key, summary in cache["issues"].items():                                                                               
      summary_lower = summary.lower()                                                                                        
      hits = sum(1 for kw in keywords_lower if kw in summary_lower)                                                          
```                                                                                                                             
  Tiny fix, eliminates 25,000 unnecessary allocations.

jwforres:
@astefanutti since your eval runs utilize create in headless mode in large scale batch, can you review this to make sure it won't have any negative impacts, take a look at the review above as well since it includes some specifics around how to handle the dup cache in headless move

astefanutti:
@jwforres thanks, you're right, that's a great use case to be covered by the eval harness. I'll test it and add on top of #9 to cover de-dup as well.

jwforres:
A few items from the earlier review that still need attention:
                                                                                
  **Must-fix**        
  ```                                                                              
  kw.lower() still called inside the inner loop (scripts/dedup_search.py:224)   
                                                                                
  hits = sum(1 for kw in keywords if kw.lower() in summary_lower)               
  ```                                                                              
  This allocates a new lowered string per-keyword per-issue. One-line fix —     
  hoist it before the loop:                                                     
  ```                                                                              
  keywords_lower = [kw.lower() for kw in keywords]
  for key, summary in cache["issues"].items():                                  
      summary_lower = summary.lower()                                           
      hits = sum(1 for kw in keywords_lower if kw in summary_lower)             
  ```                                                                              
  **Should-fix:** cache build blocks interactive UX                                 
                                                                                
  Right now the first search() call (or any call after the 4h TTL expires)      
  silently blocks while paginating through every open issue. For a 2,000-issue
  project that's ~20 sequential API calls — 6-16 seconds of unexplained dead    
  time before the user even gets asked their first clarifying question. That's a
   UX regression over the current /rfe.create flow where the conversation starts
   immediately.

  **At minimum for this PR**: print a message to stderr when starting a cache build 
  (e.g., "Building duplicate-detection cache (first run)...") so the user knows
  why there's a pause. The opt-in prompt and headless skip from the earlier     
  review can be a follow-up, but the user shouldn't be left staring at nothing.

  Can defer (but tracking)                                                      
   
  These are fine as follow-up work, just want to make sure they don't get lost: 
                  
  - Incremental cache refresh — use updated >= "<last_refresh>" JQL to fetch    
  only the delta instead of a full rebuild every 4 hours. Major improvement for
  larger projects.                                                              
  - YAML → JSON for cache format — json.load/json.dump is 10-20x faster than
  PyYAML for flat dicts and eliminates the try/except ImportError guard. Drop-in
   change.
  - _extract_fallback_keywords clarification — the SKILL.md always passes       
  --keywords, so this path is never exercised by the skill. Either mention in   
  the SKILL.md that --keywords is optional (so the fallback has a purpose), or
  document it as defensive code.
  
  If you want we can open issues for the follow up work to track.

syntaxsdev:
@jwforres Sorry missed that - good point. Fix coming now.

jwforres:
Closing — all changes are included in #93, which also adds the eval suite and three bug fixes (compound JQL, ISO dates, AskUserQuestion flow). Tracking everything there.

PR #88, opened 2026-04-14
Improve speedrun pipeline robustness

astefanutti:
## Summary

- Add post-auto-fix completeness check (`check_autofix_complete.py`) that verifies all RFEs have review files, with a retry loop for missing IDs
- Add guardrail in auto-fix dispatch loop against premature pipeline termination — Sonnet would stop early and skip to the summary phase when it estimated it was running low on context
- Harden pipeline scripts for reliability:
  - `collect_recommendations.py`: treat missing review files and parse errors as errors instead of silently skipping
  - `pipeline_state.py`: add 5-minute `parallel_timeout` for feasibility agents, enforce `post_verify` before phase advancement, execute report generation command in REPORT phase
  - `prep_assess.py`: add `--clean-all` flag to remove stale assessment files between runs, validate task file content (minimum 50 chars) before copying to assessment directory

## Context

Discovered during agent-eval-harness runs comparing Opus vs Sonnet on `rfe.speedrun` with 20 test cases. Sonnet completed Phase 1 (create 20 RFEs) but terminated during Phase 2 (auto-fix/review), producing zero review files. Analysis of the stream-json logs showed Sonnet concluded it was running low on context and emitted a text-only response to skip to the summary — which triggers the pipeline execution constraint that terminates the CI process.

These changes provide defense-in-depth: the CRITICAL guardrail in auto-fix tells the model not to make preemptive exit decisions, and the completeness check in speedrun detects and retries incomplete auto-fix runs.

## Test plan

- [ ] Re-run eval with Sonnet: `/eval-run --model sonnet`
- [ ] Verify Opus eval still passes (no regression)
- [ ] Confirm `check_autofix_complete.py` correctly detects missing reviews and outputs `MISSING_IDS=`

## Summary by CodeRabbit

* **Chores**
  * Enforced run-to-completion behavior across auto-fix and speedrun pipelines; submit now aligns with headless mode.
  * Added a completeness check and automatic retry cycle for post-auto-fix verification.
  * Improved error detection when collecting reviews and added stricter validation during assessment preparation (including a clean-all option and skipping invalid/empty tasks).
  * Phase timeouts added to improve pipeline reliability.

### DIFF
--- a/.claude/skills/rfe.speedrun/SKILL.md
+++ b/.claude/skills/rfe.speedrun/SKILL.md
@@ -2,7 +2,7 @@
 name: rfe.speedrun
 description: End-to-end RFE pipeline. Accepts a single idea, Jira key(s), or a YAML batch file. Creates, reviews, auto-fixes (with splits), and submits. Supports --headless, --announce-complete, and --dry-run for CI.
 user-invocable: true
-allowed-tools: Read, Write, Edit, Glob, Grep, Bash, AskUserQuestion
+allowed-tools: Read, Write, Edit, Glob, Grep, Bash, AskUserQuestion, Skill
 ---
 
 You are running the full RFE pipeline in speedrun mode. Your goal is to go from problem statements to submitted Jira tickets with minimal interaction. You orchestrate by calling other skills — never duplicate their work.
--- a/.claude/skills/rfe.speedrun/SKILL.md
+++ b/.claude/skills/rfe.speedrun/SKILL.md
@@ -113,7 +113,7 @@ python3 scripts/check_autofix_complete.py
 
 If incomplete (exit code 1), the output shows `MISSING_IDS=RFE-006,RFE-007,...`. Re-invoke auto-fix with only the missing IDs:
 
-```
+```text
 /rfe.auto-fix [--headless] [--batch-size N] <missing_IDs>
 ```
 
--- a/.claude/skills/rfe.speedrun/SKILL.md
+++ b/.claude/skills/rfe.speedrun/SKILL.md
@@ -2,7 +2,7 @@
 name: rfe.speedrun
 description: End-to-end RFE pipeline. Accepts a single idea, Jira key(s), or a YAML batch file. Creates, reviews, auto-fixes (with splits), and submits. Supports --headless, --announce-complete, and --dry-run for CI.
 user-invocable: true
-allowed-tools: Read, Write, Edit, Glob, Grep, Bash, AskUserQuestion, Skill
+allowed-tools: Read, Write, Edit, Glob, Grep, Bash, AskUserQuestion
 ---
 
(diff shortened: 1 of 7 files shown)


.claude/skills/rfe.speedrun/SKILL.md
```diff
@@ -2,7 +2,7 @@
 name: rfe.speedrun
 description: End-to-end RFE pipeline. Accepts a single idea, Jira key(s), or a YAML batch file. Creates, reviews, auto-fixes (with splits), and submits. Supports --headless, --announce-complete, and --dry-run for CI.
 user-invocable: true
-allowed-tools: Read, Write, Edit, Glob, Grep, Bash, AskUserQuestion, Skill
+allowed-tools: Read, Write, Edit, Glob, Grep, Bash, AskUserQuestion
```
jwforres: Doesn't this prevent speedrun from invoking the autofix skill?

astefanutti: That's a good catch, I don't think that change should have been committed. It's removed now.

astefanutti:
@jwforres I've rebased this one and addressed comments.

I've tested it successfully with #74 that's also rebased.

Note that even with Opus the auto-fix dispatch loop deviation issue happens from time to time as well so it's worth fixing in any case not just for Sonnet.

jwforres (commented):
Two issues found during review:

**1. Report runs twice in the REPORT phase**

`PHASE_CONFIG` defines REPORT as a `script` type with `command: "python3 scripts/generate_run_report.py --start-time {start_time} --batch-size {batch_size}"`. The normal script phase flow is: `next-action` returns `run_script` → LLM calls `run-phase` → command executes → dispatch marker written → LLM calls `next-action` again → marker found → `advance()` called.

The new `advance()` code for the REPORT phase also calls `generate_run_report.py`. So the report gets generated once by `run-phase` and then again by `advance()`. Either REPORT should become a `noop` type (removing the PHASE_CONFIG command and letting `advance()` handle everything), or the `advance()` code should only handle `finish.py` and the state transition.

**2. `prep_assess.py` content validation concerns**

Two issues with the 50-char minimum check:

If this is intended to catch empty placeholder files left behind by `next_rfe_id.py` (which creates 0-byte files via `open(placeholder, "a").close()`), it should check for that directly — e.g. empty file or missing frontmatter delimiters — rather than an arbitrary 50-char threshold that doesn't correspond to any actual boundary in the file format.

Separately, this check runs as `pre_script` in all four ASSESS phases (ASSESS, REASSESS, SPLIT_ASSESS, SPLIT_REASSESS), not just in the speedrun context. A failure calls `sys.exit(1)` through `_run_script`, which kills the entire `next-action` process mid-wave — remaining IDs in the wave never get their pre_scripts run and no agents launch. A single bad file shouldn't take down the whole wave; consider logging a warning and skipping the ID so the rest of the wave can proceed.

jwforres (approved):
LGTM

PR #93, opened 2026-04-23
Evaluation for the duplicate detection feature

astefanutti:
## Summary

Adds an evaluation suite for the duplicate detection feature introduced by #76 using the [agent-eval-harness](https://github.com/opendatahub-io/agent-eval-harness) with a [jira-emulator](https://github.com/jctanner/jira-emulator) as the Jira backend.

### Evaluation setup

- **eval-rfe-create.yaml**: per-case eval config that runs `/rfe.create` in interactive mode (no `--headless`) against 4 test cases, each with its own isolated workspace and jira-emulator credentials injected via `execution.env`
- **4 dedup test cases** covering the full matrix:
  - **case-021** (high-overlap duplicate): skill should detect the duplicate, confirm with user, and stop
  - **case-022** (partial keyword overlap): dedup finds no real matches, RFE created
  - **case-023** (novel topic): dedup may find false-positive matches, skill proceeds and creates RFE
  - **case-024** (false positives from keyword overlap): same — skill proceeds correctly
- **5 judges**: `rfe_outcome_correct` (annotation-aware pass/fail), `dedup_search_executed` (JQL returned valid JSON), `no_crash` (clean exit), `dedup_question_answered_correctly` (inline stdout parsing of AskUserQuestion interactions), `rfe_quality` (LLM judge, conditionally skipped for duplicate cases via `if` field)
- **answers.yaml** per case with `dedup_guidance` for the LLM-based AskUserQuestion hook (Sonnet) to answer domain-specific questions correctly
- **seed-jira-emulator.sh**: seeds the emulator with 20 realistic RFE issues sourced from eval run `2026-04-19-opus-46-v2`

### Fixes to dedup (found during eval testing)

- **Compound JQL**: `_build_text_query_jql` now emits `(text ~ "kw1" OR text ~ "kw2")` instead of Lucene OR syntax inside a single `text ~` value — portable across Jira Cloud and jira-emulator
- **ISO dates**: `_search_jql` uses `created >= "YYYY-MM-DD"` instead of `created >= -30d` — the relative date syntax is not supported by all Jira implementations
- **AskUserQuestion for matches**: Step 1.5 now presents matches and asks the user in a single AskUserQuestion call, preventing text-only responses from ending the conversation in `--print` mode

### Results

All 4 cases pass all judges at 100% with the fixes applied (run `2026-04-23-opus-v3`).

## Test plan

- [x] 59 dedup unit tests pass
[...]

PR #133, opened 2026-07-21
Add /rfe.merge design proposal and Jira capability probe

raravind007:
## Summary

`/rfe.merge` would be the inverse of `/rfe.split`: given two or more RHAIRFE tickets that are actually the same request, pick a survivor, fold the others' content into it, link and close the losers with credited provenance. This PR is the design only — no executor code — because the write path (closing real Jira tickets) is destructive enough that I wanted the phase-by-phase logic settled before writing anything that runs.

The design doc (`design-proposals/rfe-merge-design.md`) went through 8 rounds of adversarial self-review plus this repo's own CodeRabbit pass. Most of the bugs it found were the same *shape* recurring: a fingerprint check written against "the state before this phase ran" kept turning into "the state before some *earlier* phase ran" as more phases got added, and each fix in isolation looked right until the next phase exposed the gap. The decision log at the top of the doc keeps every round's finding and fix, including the ones that turned out to be wrong on a later pass (e.g. an earlier fix suggested a Jira-comment "lease" for concurrent-executor safety, which doesn't actually work — Jira has no atomic create-if-absent, so a lease built from a Jira write just relocates the race instead of closing it. Section 2 now says this plainly as a non-goal instead.)

The capability probe (`design-proposals/rfe-merge-capability-probe.py`) is the part that's actually runnable — read-only, identity-redacted by default, used to validate the design's assumptions against the real RHAIRFE instance (confirmed via JQL survey that `Duplicate` is the link convention in active use — 50+ issues vs. 0 for the JPD `Polaris merge work item link` type that superficially looked like a candidate).

## What's genuinely still open

- The one remaining `PENDING_REHEARSAL` item: confirming link orientation (does the absorbed ticket's view say "duplicates X" or "is duplicated by X") requires an actual write against a real pair, which I haven't done yet.
- Whether `Duplicate` vs. the Polaris link type is the right default beyond this one instance — resolved for RHAIRFE, not a general claim.
- No implementation exists yet. This PR is asking "is the design right," not "does the code work."

## Test plan

- [x] Probe: `py_compile`, `ruff check`, `--help`
- [x] `tests/test_rfe_merge_capability_probe.py` — 37 tests covering redaction correctness (identity fields, nested objects inside linked issues, custom field names, instance URLs), HTTPS enforcement, and error-handling paths
- [x] All 5 YAML examples in the design doc parse; Markdown fences balanced and labeled
- [ ] Manual Jira rehearsal (not yet run — the one open item above)

## Summary by CodeRabbit

* **New Features**
  * Added a read-only CLI to securely probe Jira merge capabilities and generate redacted, pretty-printed JSON reports.
  * Enforced HTTPS-only server validation and stronger output redaction to avoid leaking identity, credentials, server details, or error bodies.
* **Documentation**
  * Added a comprehensive design proposal for an approval-gated `/rfe.merge` workflow, including invariants, verification checkpoints, and recovery/resumability rules.
* **Tests**
  * Added extensive pytest coverage for redaction, summarization correctness, error handling, and CLI behavior.
* **Chores**
  * Updated version control ignore rules to exclude local diagnostic probe output (`merge-capability-probe.json`).


raravind007:
@jwforres @astefanutti — requesting review on the design itself, not just the diff mechanics.

@astefanutti, specifically: this proposal is adjacent to what #76/#93 tried to do (detecting/handling duplicate RHAIRFE tickets), and I went back through that history before writing this — the evals on #93 were passing 4/4 when it got closed, and as far as I can tell the actual reason was the unresolved security review debt around free text flowing into JQL and shell interpolation, not a design flaw in the dedup logic itself. I'd genuinely like your read on whether that's the right lesson to draw, and whether you see the same failure mode anywhere in this design — the write path here touches Jira comments/links/transitions rather than JQL construction, but if there's a structural reason this effort would hit the same wall, I'd rather find out now than after building an executor against it.

PR #143, opened 2026-07-26
Initiative Support

danmcp:
Works with: https://github.com/opendatahub-io/assess-rfe/pull/10/

## Description
Add Initiative (RHOAIENG) pipeline support with eval coverage
Extends the RFE creator pipeline to handle Jira Initiatives (RHOAIENG project) alongside the existing RFE (RHAIRFE) pipeline, and adds a dedicated evaluation dataset and CI to guard it.

Add Initiative pipeline support alongside existing RFE pipeline
Mirrors the RFE pipeline with initiative-specific skills, review types, and Jira field mappings.

**New skills**
initiative.create — build an Initiative from a free-text objective (objective/problem/scope template)
initiative.review — orchestrates scoring + feasibility review + strategic-alignment review (when a RHAISTRAT parent exists)
initiative.split — split oversized Initiatives into focused children
initiative.submit — submit to RHOAIENG as issue type "Initiative" (id 10103)
initiative.auto-fix — batch: assess → review → revise → reassess → split → collect
initiative.speedrun — end-to-end: create → auto-fix → submit → summary (batch YAML input)
initiative-feasibility-review and strategic-alignment-review agents

**Pipeline infrastructure**
PIPELINE_TYPES dispatch in pipeline_state.py routes all phase operations through item-type config (dirs, schemas, skill names, parallel agents)
Every script takes --type initiative|rfe (defaults to rfe for backward compatibility)
_save_originals() lets check_revised.py detect revisions for newly-created items, not just Jira-fetched ones
Early bootstrap (Step 0.5) in speedrun skills registers agent definitions before the CREATE phase (fixes agent-registration race)
Explicit Skill(...) invocations in speedrun skills prevent stochastic Agent selection that terminated CI sessions
Tests: 12 new test files; all existing tests updated for dual-type support

**Add initiative eval dataset and CI workflow**
16-case eval dataset (eval/initiative-dataset/) spanning process-oriented, platform-engineering, component-extraction, cross-cutting, and testing/CI initiative types — each with input.yaml + annotations.yaml
eval-initiative.yaml — 6 check judges + 2 LLM judges (initiative_quality, revision_quality)
eval/config/initiative-pairwise-judge.md — pairwise A/B comparison prompt
.github/workflows/rfe-creator-eval.yml — CI trigger for initiative eval runs
eval/README.md — dataset documentation

**Backward compatibility**
The existing RFE pipeline is unchanged in behavior — all scripts default to --type rfe. Initiative support is purely additive.

## How Has This Been Tested?
Unit and Integration Testing has been added
Also tested with Eval and the jira emulator

## Merge criteria:


danmcp:
Eval example with RFEs to check for regressions:

--- Eval Summary ---
# Evaluation Summary
**Skill:** rfe.speedrun  
**Model:** claude-opus-4-6  
**Run ID:** 2026-07-28-opus-4-6  
**Date:** 2026-07-28
## Overall Result: PASS
All check judges passed with 100% pass rate. LLM judges (revision_quality, rfe_quality) exceeded quality thresholds.
## Judge Results
| Judge | Type | Mean Score | Pass Rate | Threshold | Status |
|-------|------|------------|-----------|-----------|--------|
| architecture_context_used | check | 1.0 | 100% | 1.0 | ✅ PASS |
| files_exist | check | 1.0 | 100% | 1.0 | ✅ PASS |
| frontmatter_valid | check | 1.0 | 100% | 1.0 | ✅ PASS |
| pipeline_flow | check | 1.0 | 100% | 1.0 | ✅ PASS |
| recommendation_consistency | check | 1.0 | 100% | 1.0 | ✅ PASS |
| revision_quality | llm | 4.9 | N/A | 4.0 | ✅ PASS |
| rfe_quality | llm | 4.45 | N/A | 4.0 | ✅ PASS |
| run_report_exists | check | 1.0 | 100% | 1.0 | ✅ PASS |
## Execution Metrics
| Metric | Value |
|--------|-------|
| Test Cases | 20 |
| Duration | 60.9 minutes (3654.6s) |
| Total Cost | $53.82 |
| Cost per Case | $2.69 |
| Total Turns | 1006 |
| Avg Turns per Case | 50.3 |
| Input Tokens | 4,624 |
| Output Tokens | 503,886 |
| Cache Read | 40,498,739 |
| Cache Create | 3,351,314 |
| Cache Hit Rate | 92.3% |
| Exit Code | 0 |
## Per-Case LLM Judge Scores
| Case | revision_quality | rfe_quality |
|------|------------------|-------------|
| case-001-model-signature-verification-at-serving | 5 | 4 |
| case-002-autoscaling-support-for-elastic-ray-clus | 5 | 5 |
| case-003-experiment-tracking-integration-for-ml-w | 5 | 5 |
| case-004-command-line-interface-cli-for-ray-clust | 5 | 5 |
| case-005-responses-api-observability-integration | 5 | 5 |
| case-006-reliable-multi-turn-stateful-tool-callin | 5 | 5 |
| case-007-allow-project-and-workbench-environment | 5 | 5 |
| case-008-add-in-ability-to-registerdeploy-an-exte | 5 | 4 |
| case-009-notebook-image-preview-for-nvidia-rubin | 5 | 4 |
| case-010-llamastack-support-for-claude-code-agent | 5 | 4 |
| case-011-mlflow-prompt-template-integration-in-ge | 5 | 5 |
| case-012-mlflow-observability-bridge-for-rhoai-mo | 5 | 5 |
| case-013-update-universal-workbench-image-with-rl | 5 | 5 |
| case-014-maintain-rayjobs-via-the-codeflare-sdk | 5 | 5 |
| case-015-ga-readiness-for-the-responses-api | 5 | 5 |
| case-016-red-hat-openshift-ai-rhoai-operator-rena | 5 | 4 |
| case-017-clone-rename-llama-stack-to-remove-meta | 5 | 4 |
| case-018-autogluon-as-a-supported-model-serving-r | 5 | 4 |
| case-019-add-mlflow-integration-for-experiment-tr | 4 | 4 |
| case-020-mlflow-experiment-tracking-integration-f | 5 | 5 |
## Conclusion
The rfe.speedrun skill executed successfully with claude-opus-4-6, completing all 20 test cases in approximately 61 minutes. All structural judges passed at 100%, and both LLM quality judges exceeded their thresholds (revision_quality: 4.9/5.0, rfe_quality: 4.45/5.0). The pipeline demonstrated strong cache utilization (92.3% hit rate) and consistent execution across diverse RFE scenarios. Quality scores indicate the pipeline produces high-quality RFEs with appropriate revision decisions and robust feasibility analysis.

scripts/submit.py
```diff
@@ -675,86 +734,93 @@ def _maybe_approve(rfe_id, jira_key, entry):
...
+                    print(
+                        f"  {item_id}: Would create {cfg['project']} {cfg['issue_type']}: {title}"
+                    )
+                    results[item_id] = f"{jira_prefix}DRY"
                 else:
+                    create_kwargs = {}
+                    # Read parent_key from frontmatter for new items
+                    task_fm, _ = read_frontmatter_validated(entry["task_path"], cfg["task_schema"])
+                    pk = task_fm.get("parent_key")
+                    if pk:
```
astefanutti: **🔴 Blocking — backward-compat regression on the RFE flow.** This forwards `parent_key` to `create_issue` unconditionally. When a *newly-created* RFE (`RFE-001`) is split during auto-fix, children get `parent_key: RFE-001` (a **local** id). The local archived parent isn't routed to `split_submit.py` (split-parent detection requires the Jira prefix, line 262) and `_has_jira_ancestor` returns `False` (line 278), so the children reach this create branch and `fields.parent = {"key": "RFE-001"}` is sent to Jira → **HTTP 400** (parent doesn't exist) → `create_issue` re-raises → `submit` exits 1. `main` never forwarded `parent_key` on create, so these children were created as standalone tickets and the run succeeded. This breaks the documented create→split→submit speedrun path.

```python
pk = task_fm.get("parent_key")
if pk and not pk.startswith(cfg["local_prefix"]):
    create_kwargs["parent_key"] = pk
```
(keeps the intended initiative→RHAISTRAT linking; restores main's behavior for local split children). The one covering test runs `--dry-run`, which skips this branch — add a non-dry-run test asserting `create_issue` is called *without* `parent` for a local-parent child.

scripts/error_collect.py
```diff
@@ -134,13 +158,13 @@ def main():
...
             if os.path.exists(rc):
                 os.remove(rc)
 
         # Clean up split artifacts
         if is_split_error:
-            split_status = f"artifacts/rfe-reviews/{rfe_id}-split-status.yaml"
+            split_status = f"{tc['reviews_dir']}/{rfe_id}-split-status.yaml"
             if os.path.exists(split_status):
                 os.remove(split_status)
             # Clean children via cleanup_partial_split.py
```
astefanutti: **🔴 Blocking (initiative-only) — split-error recovery runs against the wrong artifact tree.** The `cleanup_partial_split.py` call just below (lines 171–173) omits `--type`, so it defaults to RFE mode even though `error_collect` was invoked `--type initiative` and uses the typed config (`tc[...]`) everywhere else. For an initiative split failure it scans `rfe-tasks` (never `initiatives`), finds no children, and its parent lookup is RFE-prefix-only → silent no-op: orphan child initiatives stay on disk and the parent stays `Archived`. On retry the split re-runs → **duplicate children in Jira** (the RHAIFIRST-306 partial-split hazard class), with the RFE-path safeguard silently absent for initiatives.

```python
subprocess.run(["python3", "scripts/cleanup_partial_split.py", "--type", pipeline_type, rfe_id], capture_output=True)
```
Add a test that an initiative split-error cleanup deletes orphan initiative children and un-archives the parent.

scripts/generate_review_pdf.py
```diff
@@ -10,7 +10,97 @@
...
+        "originals_dir": "initiative-originals",
+        "id_field": "initiative_id",
+        "jira_prefix": "RHOAIENG-",
+        "local_prefix": "INIT-",
+        "entity_name": "Initiative",
+        "entity_name_plural": "Initiatives",
+        "report_title": "Initiative Review &amp; Remediation Report",
+        "default_output": "initiative-review-report.html",
+        "pass_threshold": 7,
+        "criterion_keys": [
```
astefanutti, line 73: **🟠 Wrong criterion keys for initiatives.** These keys (`objective` / `problem_statement` / `success_criteria` / `timeline_dependencies`) don't match the actual initiative-review score fields — which are `what` / `why` / `scope` / `open_to_how` / `right_sized` (see `generate_run_report.py` `score_fields` and the `artifact_utils` schema). Only `scope` overlaps. Result: every initiative HTML report renders the four phantom criteria as `0`, never shows the real ones, and lists all initiatives as gaps. `submit.py` invokes this with `--type initiative`, so it's always wrong for initiative runs. Best fix: derive the keys from the one schema registry so scripts + schema + PDF can't drift (see the follow-up note in the review summary).

.claude/skills/initiative.speedrun/SKILL.md
```diff
@@ -0,0 +1,219 @@
...
+Count entries and pre-allocate all IDs upfront:
+
+```bash
+python3 scripts/next_rfe_id.py --prefix INIT --dir artifacts/initiatives --from-batch <input_file>
+```
+
+For each entry, launch an Agent to invoke `/initiative.create`. Pass the pre-assigned ID so each Agent knows which ID to use:
+
+```
+Agent for entry 1:  /initiative.create --headless --initiative-id INIT-001 [--parent <parent_key>] <prompt>
```
astefanutti: **🟠 Silent priority loss — drift from `rfe.speedrun`.** This Mode A create template omits the `[--priority <priority>]` that `rfe.speedrun:85` passes. `validate_batch_input.py` accepts and validates a per-entry `priority`, but it's never forwarded to `initiative.create`, so every batch entry with a non-default priority is submitted to RHOAIENG as `Normal`. Add `[--priority <priority>]` here (and consider `[--labels ...]`, which both speedruns currently drop) to mirror `rfe.speedrun`.

.claude/skills/initiative.review/SKILL.md
```diff
@@ -0,0 +1,245 @@
...
+
+```
+Read the skill file at .claude/skills/strategic-alignment-review/SKILL.md and follow all instructions in the body (everything after the YAML frontmatter). The Initiative ID to review is: <ID>
+```
+
+If no RHAISTRAT parent_key, skip the alignment agent for this ID.
+
+Launch all agents for all IDs in parallel (up to 3N agents total for N IDs: assess + feasibility + alignment).
+
+Wait for all to complete. After completion, check prerequisites for each ID via Glob:
```
astefanutti: **🟠 Reliability discipline lost in the fork (design drift).** `rfe.review` uses the bounded poll-barrier protocol (`check_review_progress.py` + `NEXT_POLL`) at every wave — **19 occurrences**; this forked skill has **0**, replacing them with *"Wait for all to complete."* The `max-2-cycle` reassess loop and its post-reassess regression catch are also collapsed to a single pass. This is the invariant *procedural spine* — the part that should stay identical, not the part that legitimately differs (rubric / alignment gating).

Blast radius is bounded: the headless CI pipeline runs through `pipeline_state.py` (which keeps `wait-for-wave` + `filter_for_revision`), so this degrades the **interactive `/initiative.review`** and **standalone `/initiative.split` → review** paths. Please port `rfe.review`'s poll loop and the reassess cycle here. (Exhibit A for the drift follow-up in the summary.)

scripts/submit.py
```diff
@@ -347,91 +411,108 @@ def _has_jira_ancestor(rfe_id):
...
             rec = review_data.get("recommendation", "submit")
 
-        # Collect needs-attention info for Jira comment
         original_labels = task_data.get("original_labels") or []
         attn_reason = None
         if review_data and review_data.get("needs_attention", False):
             attn_reason = review_data.get("needs_attention_reason")
 
-        # Determine if this RFE qualifies for auto-approve
         auto_approve = (
```
astefanutti, line 498: **🟠 Shared-code behavior change on the RFE path — please confirm intent + add tests.** `main`'s gate was `pass AND feasibility == "feasible"`; this is `pass AND feasibility != "infeasible" AND not needs_attention`. So an RFE with `feasibility=indeterminate` now auto-approves where it previously did not — a change to the existing pipeline (contradicts the "purely additive" claim). Live impact is low (no shipped skill passes `--auto-approve`) but it's untested. If RFE parity is intended, gate the loosening per-type in `TYPE_CONFIGS`; if intended for both, document it and fix the approve-comment wording (it claims "technical feasibility checks" passed). Either way add tests pinning each AND-term (indeterminate→approved, infeasible→not, needs_attention→not) for both types.

scripts/pipeline_state.py
```diff
@@ -630,18 +733,23 @@ def cmd_init(args):
         os.remove(DISPATCH_MARKER)
     state = {
         "phase": "INIT",
+        "type": opts.type,
         "batch": 0,
         "total_batches": 0,
         "headless": opts.headless,
         "announce_complete": opts.announce_complete,
         "batch_size": opts.batch_size,
+        "max_concurrent": opts.max_concurrent,
```
astefanutti: **🟡 Undocumented RFE throughput change.** `cmd_init` now always persists `max_concurrent` (default 6), and the wave sites read it with a `batch_size` fallback that is now dead for fresh state. On `main`, per-wave concurrency was driven by `batch_size`, so a direct `/rfe.auto-fix` over a large set (default `batch_size=50`) ran ~50 agents/wave; now it's capped at 6 (~8× fewer per wave). Output is identical, but this contradicts "RFE unchanged." If the cap is an intentional stability throttle (plausible given the concurrent-agent flake history), document it; otherwise compute `max_concurrent or batch_size` at the wave sites.

scripts/artifact_utils.py
```diff
@@ -714,6 +883,49 @@ def rename_to_jira_key(artifacts_dir, rfe_id, jira_key):
...
+    if os.path.isdir(initiatives_dir):
+        for filename in list(os.listdir(initiatives_dir)):
+            if not (filename == f"{initiative_id}.md" or filename.startswith(initiative_id + "-")):
+                continue
+            if not (filename.endswith(".md") or filename.endswith(".yaml")):
+                continue
+
+            old_path = os.path.join(initiatives_dir, filename)
+
+            if filename.endswith("-removed-context.yaml"):
```
astefanutti: **🟡 Latent clobber — missing `-comments.md` branch.** `rename_to_jira_key` (RFE) maps `-comments.md` companions before its `else`, but this function only handles `-removed-context.*`; any other `.md` companion falls to `else → {jira_key}.md`. Harmless today (the initiative fetch never writes comments), but if comment fetching is added — as it exists for RFEs — an `INIT-N-comments.md` would collide with the main file and be clobbered. Add the `-comments.md` branch, or factor the two rename functions into one shared helper.

scripts/bootstrap-assess-rfe.sh
```diff
@@ -8,16 +8,23 @@ if [ -n "${RFE_SKIP_BOOTSTRAP:-}" ]; then
...
 
+# Checkout a specific branch/tag if requested
+if [ -n "${ASSESS_RFE_REF:-}" ]; then
+  git -C "$CONTEXT_DIR" fetch origin "$ASSESS_RFE_REF" 2>&1
+  git -C "$CONTEXT_DIR" checkout FETCH_HEAD 2>&1
+fi
+
 # Validate that the rubric file exists after cloning
 if [ ! -f "$RUBRIC_FILE" ]; then
   echo "ERROR: Rubric file not found at $RUBRIC_FILE after bootstrap" >&2
```
astefanutti, line 68: **🟡 Bootstrap validates only the RFE rubric.** The initiative pipeline depends on `assess-initiative/scripts/agent_prompt.md` and the `initiative-scorer` agent (`PIPELINE_TYPES['initiative']`), but bootstrap only checks the RFE rubric before exiting 0. If upstream ever drifts, the initiative ASSESS wave launches scorers that never register/produce results → `wait-for-wave` hangs for its max wait with no actionable error (a failure mode this repo has hit). Consider validating the assess-initiative rubric + scorer here too.

astefanutti (commented):
## Review — Initiative pipeline support

Reviewed via a multi-agent pass (8 correctness/security/test/eval dimensions + an adversarial design review), with each finding re-verified against the code and `main`. The **test suite passes (650)** and the architecture is largely clean; comments below are inline. Posting as a comment (not a formal block), but I've labeled the two 🔴 items I'd treat as blocking.

### Inline findings
- 🔴 **2 blocking**: local `parent_key` forwarded to Jira on create → regresses the RFE create→split→submit flow (`submit.py`); `error_collect.py` drops `--type` into `cleanup_partial_split.py` → broken initiative split-recovery / duplicate children.
- 🟠 **should-fix**: phantom PDF criterion keys; `initiative.speedrun` drops `--priority`; `initiative.review` lost the poll-barrier + reassess cycle; shared `auto_approve` gate change (intent + tests).
- 🟡 **minor**: RFE wave-concurrency change; latent `-comments.md` clobber; bootstrap rubric validation; README 20→16 cases. ⚪ **nit**: case-005 annotation.

### Design assessment (skill duplication)
The PR makes two choices: the **backend is parameterized** (`--type`, `PIPELINE_TYPES`/`TYPE_CONFIGS`) and the **skills are fork-and-modified** (`rfe.X` → `initiative.X`). The code/prose boundary is defensible — LLM-followed prose doesn't parameterize as safely as Python, and distinct triggers are a product need. But forking the *entire* prose layer with no drift control is the weak point, and this PR is effectively a controlled experiment: **the parameterized backend didn't drift; the forked frontend did** — and it drifted in the invariant *procedural spine* (poll barriers, `--priority`, reassess cycle), not in the parts that legitimately differ (rubric, RHAISTRAT-gated alignment, split calibration). Note the parameterized layer isn't drift-*proof* either — the PDF criterion-key bug is duplicated *knowledge* in that layer. The lesson: single-source the shared knowledge and guard it.

### Follow-up (not blocking this PR)
1. **Single-source the per-type score/criterion fields** into one registry that scripts + frontmatter schema + PDF import (kills the PDF-keys drift class permanently).
2. **Add a CI drift-lint** that diffs the "supposed-to-be-shared" procedural regions across each `rfe.X`/`initiative.X` pair and fails on divergence — would have caught the barrier and `--priority` drops.
3. **Scoped shared-core extraction** of the invariant spine (poll-barrier choreography, reassess loop, the near-identical `assess-agent` prompt) via the repo's existing read-and-apply idiom; keep the judgment prose forked. Prefer pushing invariants *down* into scripts where possible (the poll barrier already is one; `batch_size` hardening could be script-enforced).
4. **Test coverage** for the initiative split path, split resume/idempotency (`discover_state`), and the `auto_approve` gate branches.
5. **Decide the boundary before a 3rd pipeline forks** (a RHAISTRAT/strategy layer looks plausible given `strategic-alignment-review`) — the duplication tax triples at that point.

eval/README.md
```diff
@@ -80,3 +80,53 @@ Input files contain only the fields the skill needs in batch Mode A — no Jira
...
+The evaluation runs the `initiative.speedrun` skill headlessly against 20 test cases derived from real RHOAIENG Jira initiatives. Each test case provides an objective (prompt + clarifying context), and the pipeline creates, reviews (with assessment, feasibility, and strategic alignment), auto-fixes, and (dry-run) submits Initiatives.
+
+### Configuration
+
+- **`eval-initiative.yaml`** — defines the skill, dataset, outputs, judges, and thresholds.
+- **`eval/config/initiative-pairwise-judge.md`** — prompt for blind A/B comparison across runs.
+
+### Dataset
+
+`eval/initiative-dataset/cases/` contains 20 test cases, each with:
```
astefanutti: **🟡 Docs mismatch.** This section says "20 test cases" (here and above) and "Two cases with empty descriptions," but the initiative dataset has **16** cases and none are empty (min `clarifying_context` ≈195 chars; only `case-012` is tagged `sparse-input`). Looks copy-pasted from the RFE section. Please correct to 16 and rewrite the sparse-input line.

eval/initiative-dataset/cases/case-005-notifications-platform/annotations.yaml
```diff
@@ -0,0 +1,17 @@
+difficulty: null
+expected_alignment: null
+expected_feasibility: null
+expected_pass: null
+expected_recommendation: split
```
astefanutti, line 5: **⚪ Nit — inconsistent reference expectation.** `expected_recommendation: split` contradicts `expected_scores.right_sized: 1`: the skill only recommends `split` when `right_sized == 0` (`review-agent.md:47`). No eval impact today (no judge reads the `expected_*` fields), but if an accuracy judge is added later this case can never match. Set `right_sized: 0` (matching the "bundles two independent efforts" known_issue) or change the recommendation to `revise`.

astefanutti:
Opened a follow-up PR that builds on this branch and addresses two things from the review above: #150

- **Single-source review criteria** (`scripts/review_criteria.py`) — the schema, run report, and HTML report now derive score fields/labels from one registry, which fixes the phantom initiative criterion keys flagged inline on `generate_review_pdf.py` (reports were rendering zeros for the real `what/why/scope/open_to_how/right_sized`). RFE behavior unchanged.
- **Skill-parity drift lint** (`scripts/check_skill_parity.py`, wired into the `Lint` workflow) — guards the fork-drift class: it fails CI when an invariant present in one of a `rfe.X`/`initiative.X` pair is missing from the other. The three currently-known drifts (poll barrier, reassess cycle, `--priority`) are baselined in `scripts/skill_parity_baseline.json` so it lands green and ratchets forward.

The three baselined entries map 1:1 to the inline comments here — as those get fixed in this PR, deleting the matching baseline id locks the fix in. It's stacked on top of `initiative_support`, so merge this PR first and I'll rebase #150 onto `main`.

scripts/submit.py
danmcp: Good catch, and the trace is exactly right. Fixed in `8384720` with your suggested diff verbatim, plus one addition:

```python
pk = task_fm.get("parent_key")
if pk and not pk.startswith(cfg["local_prefix"]):
    create_kwargs["parent_key"] = pk
elif pk:
    print(f"  {item_id}: Parent {pk} is not in Jira — creating without parent")
```

The `elif` is there so the drop shows up in the run log. Silently discarding a `parent_key` is the kind of thing that looks like a linking bug six months later, and this way the speedrun output says why the child is standalone.

Agreed on `local_prefix` over `jira_prefix`, and the initiative path makes it more than a stylistic call — an Initiative's parent is legitimately a `RHAISTRAT-` Outcome key, which is neither the local prefix nor `cfg["jira_prefix"]`. An allowlist test would drop those. I left a comment at the call site saying so, since "shouldn't this check `jira_prefix`?" is the obvious later "cleanup".

You were also right about why this slipped: the only covering test used `--dry-run`, which returns before the branch. Added non-dry-run coverage on both paths, running the real create against the Jira emulator:

- `test_submit_integration.py::test_children_of_local_parent_created_without_parent` — archived local `RFE-001` with two children; asserts exit 0, both issues created, `fields.parent is None` on each, and the new log line present. This is the regression as you described it.
- `test_initiative_submit_integration.py::test_local_parent_is_not_forwarded` — same shape for `INIT-` parents.
- `test_initiative_submit_integration.py::test_strategic_outcome_parent_is_forwarded` — the positive case: `RHAISTRAT-1234` must still arrive as `fields.parent`, and `"is not in Jira"` must *not* appear. Without this one, an over-broad exclusion would pass the other two.

scripts/error_collect.py
danmcp: Confirmed and fixed in `8384720`. `--type` is now threaded through:

```python
# Clean children via cleanup_partial_split.py. Without --type it
# defaults to rfe and would scan rfe-tasks/ for an initiative's
# children, finding none and leaving them orphaned.
subprocess.run(
    ["python3", "scripts/cleanup_partial_split.py", rfe_id, "--type", pipeline_type],
    capture_output=True,
)
```

Your read of the failure mode is what the tests confirm — silent no-op, orphaned children, parent stuck `Archived`, and then a retry that re-splits a parent which still has children. The RHAIFIRST-306 framing is apt: the safeguard existed, it just wasn't reachable from the initiative path.

Also added the test you asked for. `error_collect.py` had no test file at all, so this is `tests/test_error_collect.py` — 15 tests, run as a **subprocess** against a real workspace rather than with mocks, specifically because the thing that broke was the `--type` handoff *between* processes. Mocking `subprocess.run` would have asserted the bug into place.

Directly on this finding:

- `test_initiative_children_deleted` — orphan children removed from the initiatives tree
- `test_initiative_parent_unarchived` — parent goes back to `Ready`
- `test_initiative_split_status_deleted`
- `test_rfe_children_deleted` — regression guard, the default path unchanged
- `test_rfe_tree_untouched_on_initiative_run` — plants a decoy `artifacts/rfe-reviews/RHOAIENG-50-review.md` and asserts it survives, so the test fails if the dirs stop being type-selected rather than just happening to be empty

Negative-checked by removing the flag again: the two child-cleanup tests fail.

scripts/generate_review_pdf.py
danmcp: Correct on all counts — only `scope` overlapped, and since `submit.py` always passes `--type initiative`, every initiative HTML report rendered five zeros and flagged the whole run as gaps. Fixed in `8384720`: `criterion_keys` is now `what/why/scope/open_to_how/right_sized`, with `criterion_labels`, `criterion_short_labels`, and `before_score_name_map` brought along (the last one matters — the revision-history parser looks up display names in it, and an unmapped name is dropped silently rather than erroring).

On the suggested fix: I pinned rather than refactored in this PR, on the reasoning that a real single-source extraction touches three scripts on the shared RFE path for no behavior change, which is poor value under a "RFE unchanged" contract. So `8384720` adds `tests/test_score_field_registry.py` (15 tests) instead, enforcing agreement across all four restatements — the schema's `scores` and `before_scores`, `score_fields` in `generate_run_report.py`, and `criterion_keys` here — plus: both label maps must cover every criterion, every `before_score_name_map` value must resolve to a real criterion, every `PIPELINE_TYPES` entry must have every registry, and the two types must not share a field set, so the parametrized tests can't pass on a copy-pasted registry.

One deliberate addition: the agreed names are also pinned *literally*, not just against each other. All four sources can be renamed in lockstep and still be wrong if the rubric in `assess-rfe` wasn't renamed with them, and that cross-repo edge is exactly where this class of bug lives. Negative-checked by reverting the key list: four tests fail.

Having seen #150, your version is better and I'd rather take it than keep mine — `review_criteria.py` removes the duplication instead of guarding it, which kills the class permanently as you originally said. Suggest we land it as the real fix and keep the registry test on top, retargeted at the single source: it still earns its place for the literal-name pin and the cross-map coverage checks, which a single registry doesn't give you by itself.

Not a problem with #150, just a merge heads-up since it's stacked on this branch: I've since renamed the initiative skills from dotted to dashed (`initiative.review` → `initiative-review`, etc.) in `e0081cd`, so `check_skill_parity.py`'s pair baseline will need the initiative side updated. RFE skills are unchanged, so the pairs now read `rfe.X` / `initiative-X`.

.claude/skills/initiative.speedrun/SKILL.md
danmcp: Fixed in `8384720` — the three create templates now read:

```
Agent for entry 1:  /initiative-create --headless --initiative-id INIT-001 [--priority <priority>] [--parent <parent_key>] <prompt>
```

which puts it structurally in line with `rfe.speedrun:85`, modulo `--parent`, which is a legitimate initiative-only addition.

Worth noting this is a clean instance of your broader diagnosis: `validate_batch_input.py` validates `priority` identically for both types, so the parameterized layer was fine — the flag was lost purely in forked prose, in the invariant part rather than the part that should differ.

On `--labels`: deliberately not fixed, because it isn't fork drift — it's broken on the RFE path too. Traced end to end:

- `validate_batch_input.py` accepts and validates a `labels` list for both types
- neither speedrun forwards it
- both create skills declare `--labels` at line 15 and never mention it again
- neither task schema has a `labels` field — only `original_labels`, which holds labels fetched *from* Jira
- `submit.py`'s `_build_labels` composes labels purely from automation state, with no user-supplied channel

So it dies at every stage on both paths. `rfe.speedrun:63` even shows `labels: [candidate-3.5]` in its batch-input example, so the RFE path advertises the feature and silently drops it — which is arguably worse than the initiative path, where the drop at least isn't documented.

Making it work needs a schema field plus submit plumbing on both types. That's a small feature, not a prose passthrough, and adding it here would mean shipping an RFE behavior change inside a PR whose contract is "RFE unchanged" — so I'd rather it get its own change. Happy to open an issue for it if you want it tracked, or fold it in here if you'd prefer it not linger.

.claude/skills/initiative.review/SKILL.md
danmcp: Accepted — this was the clearest case of the drift you're describing, and "invariant procedural spine" is the right frame. Counting agent completions isn't a completion signal; it drifts the moment one agent dies. Restored in `8384720`.

The house barrier idiom (`state.py write-ids <poll-file>` → `check_review_progress.py --phase <phase> --id-file`, sleeping the reported `NEXT_POLL`) is now applied at every wave: fetch, assess, feasibility, alignment, review, revise, and both reassess waves — eight call sites against seven in `rfe.review`, the extra one being the initiative-only alignment phase.

The two-cycle reassess loop is back with the disk-backed counter, initialized via `set-default` specifically so context compression re-entering the step can't reset it to 0 and restart the loop:

```
python3 scripts/state.py set-default tmp/initiative-review-config.yaml reassess_cycle=0
```

with the `>= 2` stop and the follow-on regression check restored.

One place I diverged from a straight port. Alignment can't use the same blocking barrier: `check_review_progress.py` exits 2 on an empty ID list, so when no ID in the batch has a RHAISTRAT parent there's nothing to poll and a naive port would fail the step. Alignment polling is therefore skipped entirely in that case, and when it does run it stops after 5 minutes rather than blocking — alignment is informational, and the prerequisite check downstream records the missing file without failing the ID. Everything else is a faithful port.

Prose-parity tests added in `test_check_review_progress.py`, negative-checked by deleting two poll lines. Agreed those are a stopgap next to a real region-differ — #150's parity lint is the better mechanism.

Two things left as-is, deliberately:

**Step 4e still says a bare "Wait for all to complete"** — `rfe.review:280` and `initiative-review:307`. This one is pre-existing and identical on both sides, so it isn't drift, and it's the revise wave rather than a wave the barrier protocol currently covers. Fixing it means editing both files together, which I'd rather do as its own change than fold into an initiative PR. Flagging it explicitly so the parity lint doesn't get baselined against a state we've already agreed is wrong.

**The `rfe.speedrun` Phase 1 disk barrier isn't ported here.** It needs a new `initiative-create` key in `PHASE_CHECKS`, and the RFE counterpart hasn't landed on this branch yet — so porting now would mean the initiative path leads the RFE path on a shared mechanism.

scripts/pipeline_state.py
danmcp: Agreed, and thanks for quantifying it — the `max_concurrent or batch_size` fallback being dead for fresh state is exactly right, which made the throttle unconditional rather than opt-in.

Fixed in `8384720` by taking the first option rather than documenting the cap: `--max-concurrent` and the `max_concurrent` state key are removed entirely (no references left anywhere in `scripts/`, `tests/`, `.claude/`, or `docs/`), so wave size derives from `batch_size` again. The concurrent-agent flake history is real, but a silent 8× throughput cut on the RFE path isn't the place to encode it — if we want a stability throttle it should be a deliberate, documented knob, and it shouldn't ship inside an initiative PR.

While removing it I extracted the duplicated computation into `_wave_size(state, config)`, since it appeared at both wave sites and had to stay in sync:

```python
batch_size = int(state.get("batch_size", 50))
n_parallel = len(config.get("parallel", []))
return max(1, -(-batch_size // (1 + n_parallel)))
```

One behavior change beyond the revert, which I want to be explicit about since this PR's contract is "RFE unchanged" and this touches the RFE path. The division now rounds **up**; it previously floored. Flooring stranded a partial slot on every batch that isn't an exact multiple of the divisor, which is what made small batches crawl — a speedrun batch of 5 ran 2 IDs per wave for RFEs and 1 for initiatives, so a 5-item batch took 3 waves and 5 waves respectively. Rounding up costs at most `n_parallel` agents over budget and keeps a small batch to a single wave far more often.

So this is a modest RFE throughput *increase* at small batch sizes, not a restoration to the byte. I judged that worth doing while the code was open and the reasoning was fresh, but it's a separable change — say the word and I'll floor it back and take the rounding fix separately.

Coverage in `test_pipeline_state.py` pins the arithmetic directly, including the rounding boundary (`batch_size=5` with 1 parallel → 3, with 2 → 2), the default (`batch_size=50` → 25 and 17), the no-parallel case, and the `batch_size=0` floor of 1.

scripts/artifact_utils.py
danmcp: Fixed in `8384720` — took the targeted branch rather than the shared-helper refactor:

```python
# Companions must be matched before the fallback: anything that
# reaches the else branch is renamed to {jira_key}.md, which is
# also the main task file's new name, so an unhandled companion
# silently overwrites it.
if filename.endswith("-comments.md"):
    new_name = f"{jira_key}-comments.md"
elif filename.endswith("-removed-context.yaml"):
    ...
```

The comment is there because the ordering is the load-bearing part and it isn't obvious from reading the branches — the failure mode isn't "companion gets a wrong name", it's "companion destroys the task file", and whichever `os.listdir` order wins decides whether you lose the companion or the submitted task.

Worth noting it's slightly less latent than "harmless today". Because the fallback target collides with the main file rather than just being wrong, any future `.md` companion in that directory hits it — not only comments. The guard is really "unhandled companion" and comments are just the first likely instance.

Two tests in `test_artifact_utils.py`: `test_comments_companion_keeps_its_suffix` asserts the companion lands at `RHOAIENG-1234-comments.md` **with its original contents**, and `test_task_file_survives_companion_rename` asserts the task file still exists at `RHOAIENG-1234.md` with `initiative_id` updated and `status: Submitted`. The second is the one that actually catches the clobber — the first can pass while the task file is gone.

On the shared-helper option: agreed it's the better end state, but folding the RFE rename into a shared function means changing behavior on the RFE submit path for no functional gain, which I'd rather not do under a "RFE unchanged" contract. Reasonable candidate for the shared-core extraction follow-up — the two functions differ only in directory names and the id field, so they'd parameterize cleanly, unlike the prose layer.

scripts/bootstrap-assess-rfe.sh
danmcp: Agreed — fixed in `8384720`.

The approach is that the caller declares which pipeline's assets it needs, rather than the script guessing: `bootstrap-assess-rfe.sh [--type rfe|initiative]`, default `rfe`, unknown arg or unknown type exits 2. An RFE run is unaffected and never looks for initiative assets.

Two gates rather than one, because the agent install is itself conditional on `$CONTEXT_DIR/agents` existing — a single post-clone check would miss the case where the clone succeeds but the agent never gets copied:

- initiative rubric checked after the clone, before the skill copy
- `initiative-scorer` checked after the agent install

Both messages name `ASSESS_RFE_REPO`/`ASSESS_RFE_REF` as the remedy, and the scorer one spells out the causal chain ("launches with `subagent_type: initiative-scorer`; without it the ASSESS phase never completes"), so a misconfigured checkout is diagnosable from the log instead of surfacing as the stall you described. `RFE_SKIP_BOOTSTRAP` still short-circuits ahead of both, so offline runs stay possible.

Callers updated: the `pipeline_state.py` SETUP phase (now type-aware), plus `initiative-review`, `initiative-speedrun`, `initiative-auto-fix`, and `initiative-split/prompts/split-agent.md`. `.claude/settings.json` also needed a wildcard allowlist entry — the existing one was an exact match, so adding an argument would have prompted for permission and stalled headless runs. That one would have been an unpleasant thing to discover in CI.

New `tests/test_bootstrap_assess.py`, 14 tests in three groups:

- **`TestTypeValidation`** — runs the real script against a fake checkout: an RFE run passes without initiative assets present, initiative fails on missing rubric, initiative fails on missing scorer, initiative passes on a full checkout, `--type=x` form accepted, unknown type and unknown argument both rejected, `RFE_SKIP_BOOTSTRAP` short-circuits.
- **`TestPathsMatchPipelineRegistry`** — pins the shell paths against `PIPELINE_TYPES`, so the two can't drift.
- **`TestCallersDeclareType`** — walks `.claude/skills/**/*.md` and asserts every initiative skill invoking bootstrap declares `--type initiative`, with a companion test guarding the filter from passing vacuously. Negative-checked by reverting one caller: the test names the file and the line.

**One deviation.** I dropped the "warn when no `--type` is declared" half of what I'd proposed. It would fire on every ordinary `rfe.review` run — noise about something that isn't wrong, which trains people to ignore it. `TestCallersDeclareType` catches the same class of mistake at CI time instead of in a log nobody reads.

For context on the env vars: the initiative rubric and `initiative-scorer` currently come from `danmcp/assess-rfe @ initiative_support`, which is the intended source for now and lands alongside this work. The gates just make the requirement explicit rather than implicit.

eval/README.md
danmcp: It was copy pasted and I originally have 20 RFE cases. I reduced the number to get the time in line due to extra initiative steps but missed the comment. Fixed in `8384720`.

Both counts in the initiative section now say 16, and the sparse-input line reads:

> One case (`case-012`, tagged `sparse-input`) provides minimal context to test sparse-input handling.

which matches the dataset — `case-012-ambient-capabilities-second-set` is the only case carrying the tag, and no case has an empty description. The RFE section's 20/two-empty claims are left alone, since they're accurate there.

Worth noting the underlying reason this drifted silently: nothing ties either count to the directory. Both are prose. A cheap guard would be a test asserting each README count matches `len(os.listdir(...))` for its dataset — happy to add it here if you want it, though it may fit better alongside the other drift guards in #150 than bolted onto this PR.

eval/initiative-dataset/cases/case-005-notifications-platform/annotations.yaml
danmcp: Right that the file was inconsistent, and thanks for catching it!

I've taken your second option instead — `right_sized` stays `1` and `expected_recommendation` becomes `revise`. That resolves the inconsistency you flagged, since `split` requires `right_sized == 0`, and it now agrees with the rubric rather than just being internally coherent.

Two neighboring fields carried the same stale assumption and are updated with it: `known_issues` said "Should split into 2", and the case was tagged `split-candidate`. Under the current framing splitting is reserved for grab-bags of genuinely separate initiatives; a 2-way bundle gets its scope narrowed by revision.

Agreed on the impact assessment — nothing reads `expected_*` today. But that's exactly why it drifted, and it's the ground truth an accuracy judge would be built against, so it's worth having right before that judge exists rather than after.

danmcp:
@astefanutti Really appreciate the review. I've updated the PR addressing the feedback. Hopefully it's in a better state now. I'm going to run a few tests and will report back with results.

danmcp:
This above run is with this PR and eval.yaml testing RFEs with https://github.com/opendatahub-io/assess-rfe

The below run is with this PE + eval-initiative.yaml testing initiatives with https://github.com/danmcp/assess-rfe/tree/initiative_support 

# Evaluation Summary
**Skill:** initiative-speedrun  
**Model:** claude-opus-4-6  
**Run ID:** 2026-08-05-opus-4-6  
**Date:** 2026-08-05
## Overall Result: **PASS**
All judge thresholds met. The pipeline successfully processed 16 test cases through the full Initiative creation, review, auto-fix, and submission workflow.
## Judge Results
| files_exist | check | 1.0 | ≥ 1.0 | ✅ PASS |
| frontmatter_valid | check | 1.0 | ≥ 1.0 | ✅ PASS |
| run_report_exists | check | 1.0 | ≥ 1.0 | ✅ PASS |
| recommendation_consistency | check | 1.0 | ≥ 1.0 | ✅ PASS |
| pipeline_flow | check | 1.0 | ≥ 1.0 | ✅ PASS |
| architecture_context_used | check | 1.0 | ≥ 0.85 | ✅ PASS |
| initiative_quality | LLM | 4.31 | ≥ 3.5 | ✅ PASS |
| revision_quality | LLM | 4.63 | ≥ 3.5 | ✅ PASS |
## Execution Metrics
| Metric | Value |
|--------|-------|
| Test Cases | 16 |
| Duration | 2,855 s (47.6 min) |
| Total Cost | $33.93 |
| Total Turns | 671 |
| Input Tokens | 5,320 |
| Output Tokens | 304,894 |
| Cache Reads | 26,944,840 |
| Cache Creates | 2,048,621 |
| Cache Hit Rate | 99.98% |
| Model | claude-opus-4-6 |
| Execution Mode | batch |
| Exit Code | 0 |
## Per-Case LLM Judge Scores
| Case | Initiative Quality | Revision Quality |
| case-001-outreach-initiatives | 4 | 4 |
| case-002-collect-early-feedback | 4 | 4 |
| case-005-notifications-platform | 5 | 5 |
| case-006-homepage-customization | 4 | 5 |
| case-009-model-registry-federation | 5 | 5 |
| case-010-pipeline-visualization-common | 4 | 5 |
| case-011-add-workflows-api-sdk | 4 | 4 |
| case-012-ambient-capabilities-second-set | 4 | 5 |
| case-013-scaleup-test-matrix-dynamic | 4 | 4 |
| case-014-shift-left-model-serving-testing | 4 | 5 |
| **Mean** | **4.31** | **4.63** |
## Conclusion
The evaluation demonstrates strong end-to-end performance of the initiative-speedrun pipeline running on claude-opus-4-6. All structural quality checks passed at 100%, architecture context integration was used consistently across all cases, and LLM judges scored both Initiative quality (4.31/5.0) and revision quality (4.63/5.0) above thresholds. The pipeline successfully produced well-calibrated Initiative assessments with accurate rubric scoring and meaningful feasibility reviews. Cache efficiency was excellent at 92.9%, enabling cost-effective execution at $33.93 for 16 full-pipeline test cases.

danmcp:
After latest changes:

# Evaluation Summary
**Skill:** initiative-speedrun  
**Model:** claude-opus-4-6  
**Run ID:** 2026-08-06-opus-4-6
## Overall Result: PASS
All judge thresholds met. The pipeline successfully created, reviewed, auto-fixed, and prepared 16 Initiatives for submission with high quality across all dimensions.
## Judge Results
| files_exist | check | 1.0 pass rate | ≥1.0 | ✓ PASS |
| frontmatter_valid | check | 1.0 pass rate | ≥1.0 | ✓ PASS |
| run_report_exists | check | 1.0 pass rate | ≥1.0 | ✓ PASS |
| recommendation_consistency | check | 1.0 pass rate | ≥1.0 | ✓ PASS |
| pipeline_flow | check | 1.0 pass rate | ≥1.0 | ✓ PASS |
| architecture_context_used | check | 1.0 pass rate | ≥0.85 | ✓ PASS |
| initiative_quality | llm | 4.375 mean | ≥3.5 | ✓ PASS |
| revision_quality | llm | 4.875 mean | ≥3.5 | ✓ PASS |
## Execution Metrics
| Metric | Value |
|--------|-------|
| Cases | 16 |
| Duration | 3146.9s (52.4 min) |
| Total Cost | $34.36 |
| Turns | 690 |
| Cache Hit Rate | 93.6% |
| Cost per Turn | $0.0498 |
| Cost per MTok | $1.1222 |
| Output Tokens per Turn | 466.26 |
## Per-Case LLM Scores
| Case | Initiative Quality | Revision Quality |
| case-001-outreach-initiatives | 5 | 5 |
| case-007-gpu-utilization-observability | 4 | 5 |
| case-011-add-workflows-api-sdk | 4 | 5 |
| case-012-ambient-capabilities-second-set | 4 | 5 |
| case-013-scaleup-test-matrix-dynamic | 4 | 5 |
| case-016-developer-portal | 4 | 5 |
**Initiative Quality Mean:** 4.375 (Good to Exemplary)  
**Revision Quality Mean:** 4.875 (Excellent)
## Conclusion
The pipeline demonstrated strong end-to-end performance across all 16 test cases. All check-based judges achieved perfect pass rates, indicating correct execution flow, valid output structure, and consistent recommendation logic. The LLM quality judges scored well above thresholds (4.375 for initiative quality, 4.875 for revision quality), showing that the pipeline produces high-quality Initiatives with accurate rubric scoring, thorough feasibility analysis, and appropriate revision decisions. The high cache hit rate (93.6%) and reasonable per-turn cost ($0.0498) indicate efficient execution. All cases correctly leveraged architecture context during feasibility reviews, and the pipeline consistently made sound decisions about when to auto-revise versus when to submit high-quality Initiatives unchanged.

astefanutti:
Re-reviewed after the addressed feedback — all 11 earlier findings are resolved (thanks, the fixes and added tests look solid), and I did a focused backward-compat pass on the existing RFE pipeline. The `rfe.*` skills are untouched except the intentional `rfe.speedrun` hardening, and the shared-script `--type rfe` default path is clean except for **one regression I'd suggest fixing before merge** (it also affects the initiative path).

### 🟠 `verify_phase.py` no longer removes failed IDs for `fetch`/`assess` — a fetch-agent failure now crashes the pipeline

[`verify_phase.py:105`](https://github.com/opendatahub-io/rfe-creator/blob/785c51e8fa932d24f361395914cf4f4a0e75ed5c/scripts/verify_phase.py#L105) gates the error-frontmatter write **and the active-set removal** behind `if failed and phase == "review":`. On `main` (c8a2a1e) `verify()` did both unconditionally for **every** phase (fetch/assess/review), so a failed ID was error-marked and dropped from `tmp/pipeline-active-ids.txt`.

FETCH still runs `post_verify --phase fetch` ([`pipeline_state.py:220`](https://github.com/opendatahub-io/rfe-creator/blob/785c51e8fa932d24f361395914cf4f4a0e75ed5c/scripts/pipeline_state.py#L220)), but it no longer removes anything, so a fetch failure leaves a stale ID with no task file. Concrete chain (verified against the code):

1. fetch agent fails to write `artifacts/rfe-tasks/{id}.md` → `verify --phase fetch` just prints `FAILED=`, ID stays in the active set.
2. `cmd_next_action`'s pre-filter keeps it (`check_id("fetch"/"assess", id) == "pending"`), so it enters the ASSESS wave.
3. the ASSESS pre-script `python3 scripts/prep_assess.py {ID}` hits the missing task file → `sys.exit(1)`.
4. `_run_script` sees the nonzero exit → `sys.exit(1)` → `next-action` (the pipeline) crashes.

On `main` this same fetch failure was tolerated (ID marked `needs_attention`/error, pipeline continued to DONE). So this converts graceful degradation into a hard pipeline crash.

**Scope:** it only triggers when a fetch/assess *agent* fails — normal runs are unaffected and the full suite (719) passes — so it's an error-path resilience regression, not a happy-path break. But given the pipeline's history with agent flakiness/barrier hangs, I'd rather not lose the graceful-degrade path. The assess-failure variant is milder (the ID survives into REVIEW where the `review`-gated branch still catches it) but wastes a review wave and mislabels the error as `review_failed`.

**Suggested fix:** restore the write+remove for `fetch`/`assess` (not just `review`), or make `prep_assess.py` / the `cmd_next_action` pre-filter skip an ID whose task file is absent. Either way worth a `test_verify_phase.py` — there isn't one today, which is why this slipped through.

### The rest of the RFE path checks out
For completeness: the previously-flagged RFE regressions (auto_approve gate, wave concurrency) are correctly reverted to `main`. The remaining `--type rfe` diffs I found are all non-regressions — intended fixes (`_save_originals` making reassess detection work for newly-*created* RFEs; `error_collect` retry-batch idempotency; wave-size floor→ceil) or cosmetic (PDF criterion label text, run-report YAML key order, richer `batch_summary` lines). One minor edge case: `generate_run_report.py` now exits 2 (no file) when `rfe-reviews` is empty/absent where `main` wrote an empty report — only matters for a degenerate empty run.

astefanutti:
> @astefanutti Really appreciate the review. I've updated the PR addressing the feedback. Hopefully it's in a better state now. I'm going to run a few tests and will report back with results.

@danmcp thanks, I've left one comment about `verify_phase.py`.

Also the last point about the "public" API and the labels like `initiative-creator-auto-created`. I'd be inclined to go with `<type>-auto-created` from now on, .e.g. `initiative-auto-created`, WDYT?

danmcp:
@astefanutti 
> Re-reviewed after the addressed feedback — all 11 earlier findings are resolved (thanks, the fixes and added tests look solid), and I did a focused backward-compat pass on the existing RFE pipeline. The `rfe.*` skills are untouched except the intentional `rfe.speedrun` hardening, and the shared-script `--type rfe` default path is clean except for **one regression I'd suggest fixing before merge** (it also affects the initiative path).

Both addressed — thanks, the second one was a real behavior change I hadn't intended.

## What changed

**`verify_phase.py`** — `if failed and phase == "review":` is `if failed:` again, so a missing output writes error frontmatter and drops the ID from the active set in every phase, same as `main`. The only remaining diff from `main` there is the type parameterization, so an initiative failure lands in `artifacts/initiative-reviews/` with `initiative_id` and `scores.scope`.

The module had no tests; added `tests/test_verify_phase.py` (33 cases) covering the failure paths for both types, plus a parametrized check that `_RFE_PHASE_OUTPUT` / `_INITIATIVE_PHASE_OUTPUT` resolve to the same paths as `check_review_progress.PHASE_CHECKS` — the invariant the barrier rests on, previously uncovered.

**`generate_run_report.py`** — half right, and the right half is worth fixing. Ran both versions:

| reviews dir | `main` | branch, before | branch, now |
|---|---|---|---|
| present, empty | exit 0, zero-count report | exit 2, no file | exit 0, zero-count report |
| absent | `FileNotFoundError`, exit 1 | exit 2, clean message | exit 2, clean message |

`main` didn't write an empty report for an absent directory — it called `os.listdir()` unguarded and crashed, so I kept the `isdir` guard. What I changed is that a present-but-empty directory falls through to `build_report([])` again, emitting YAML byte-identical to `main`'s apart from the timestamp. This matters more than "degenerate": REPORT is a `script` phase and `cmd_run_phase` does `sys.exit(result.returncode)`, so exit 2 fails the run at its terminal step — and a scheduled `/rfe.auto-fix --jql` matching nothing is a normal outcome, not a red pipeline.

## What the diffs were trying to do

Both came from the type-parameterization commit, and in both cases the intent was narrower than the effect.

`verify_phase.py`: the thought was "don't fabricate a `-review.md` for a phase that isn't the review phase." Gating on `phase == "review"` was the wrong instrument — it also switched off the active-set removal, which is the part that matters. If the artifact shape is the real objection it should be argued on its own terms; as it stands that error file is how a failure becomes visible to `collect_recommendations` and the run report.

`generate_run_report.py`: the `parser.error` was aimed at genuine misuse — wrong `--artifacts-dir`, no IDs anywhere — and I didn't separate that from a real run that legitimately reviewed zero items. Those are now distinct.

## What's still open

**A phase whose output never appears doesn't crash the pipeline — it wedges it.** On `main` too, which is why I'm raising it rather than expanding the PR. And I might be missing something still.

Your chain is fetch fails → ID stays in the active set → enters ASSESS → `prep_assess.py` exits 1 → pipeline exits. Step 3 on is right, but I don't think the ID reaches ASSESS: `post_verify` and `advance` both live inside the `if not remaining:` branch of `cmd_next_action` (`pipeline_state.py:952`), and `remaining` is built from `check_id(...) == "pending"` — the same `os.path.exists` test `verify_phase` calls a failure. `cmd_advance` has the matching guard via `_check_agent_phase_complete` (`:1093`).

So `post_verify` only runs once every output already exists, which means `verify --phase fetch` structurally can't emit a non-empty `FAILED=` through the dispatcher — the branch we just restored is currently unreachable for fetch/assess. (`review` is the exception: `check_id` calls an unparsable review file `"error"`, not `"pending"`, so the filter clears and `verify` does act.)

What happens instead: the fetch agent writes nothing on failure, the ID stays `pending`, `next-action` re-emits a FETCH wave, `wait-for-wave` returns 3 with "re-run this command", repeat. No attempt cap anywhere — `MAX_NEXT_ACTION_ITERATIONS = 50` bounds phase-hopping inside one `next-action` call, not wave retries. The run spins until CI times out.

Scope: not reachable from the initiative eval path. Reproducing it needs `/rfe.review` or `/initiative-review` against real Jira with a deterministically failing fetch, identical on `main`.

Fix shapes: track per-ID attempts and have `cmd_next_action` fail an ID past a threshold instead of re-waving it (which is also what makes `verify_phase`'s fetch/assess branch reachable — the two are complements), or give `wait-for-wave` a cumulative per-phase budget so exit 3 eventually terminates naming the stuck IDs.

If this logic is correct, I'm happy to open a an issue or PR separately.

**Smaller residual:** `main` had `choices=list(PHASE_OUTPUT.keys())` on `--phase`; with two maps this branch dropped it, so a bad phase gets `verify()`'s `Unknown phase: X` → exit 1 instead of argparse's usage error → exit 2. Can restore it as the union of both maps if you want, though that would accept `alignment` under `--type rfe` and let `verify()` reject it, which is why I left it.

danmcp:
> Also the last point about the "public" API and the labels like `initiative-creator-auto-created`. I'd be inclined to go with `<type>-auto-created` from now on, .e.g. `initiative-auto-created`, WDYT?

Agreed — initiative-creator names a tool that doesn't exist and the new names look better; the rfe-creator-* prefix only reads correctly because the tool name and item type happen to coincide there. Renamed all twelve to initiative-* (initiative-auto-created, initiative-feasibility-pass, initiative-alignment-strong, initiative-ignore, …).

One concern with the shorter prefix: split_submit.py decides which of a parent's labels are pipeline-owned with label.startswith(f"{label_prefix}-"). At initiative-creator- a collision was implausible; at initiative- a human's initiative-q3 or initiative-planning would now be treated as ours and dropped from what split children inherit. We could also swap the prefix test for the explicit list of suffixes we actually write.

danmcp:
Ran a quick mode test (2 cases) to validate the workflow is ok with the latest changes:

# Evaluation Summary

**Skill**: `rfe.speedrun`  
**Model**: `claude-opus-4-6`  
**Run ID**: `2026-08-06-opus-4-6`  
**PR**: 143

## Overall Result

**PASS** — All judges met their thresholds.

## Judge Results

| architecture_context_used | check | 1.0 (100% pass) | n/a | ✓ PASS |
| files_exist | check | 1.0 (100% pass) | n/a | ✓ PASS |
| frontmatter_valid | check | 1.0 (100% pass) | n/a | ✓ PASS |
| pipeline_flow | check | 1.0 (100% pass) | n/a | ✓ PASS |
| recommendation_consistency | check | 1.0 (100% pass) | n/a | ✓ PASS |
| run_report_exists | check | 1.0 (100% pass) | n/a | ✓ PASS |
| revision_quality | llm | 4.5 / 5 | n/a | ✓ PASS |
| rfe_quality | llm | 4.0 / 5 | n/a | ✓ PASS |

## Execution Metrics

| Metric | Value |
|--------|-------|
| Cases | 2 |
| Duration | 686.6 s (11.4 min) |
| Cost | $7.38 |
| Turns | 174 |
| Cache Hit Rate | 0.942758 (94.3%) |
| Cost per Turn | $0.042389 |
| Cost per MTok | $0.9474 |
| Output Tokens per Turn | 218.94 |

## Per-Case LLM Scores

| Case | revision_quality | rfe_quality |
|------|------------------|-------------|
| case-001-model-signature-verification-at-serving | 5 / 5 | 4 / 5 |
| case-002-autoscaling-support-for-elastic-ray-clus | 4 / 5 | 4 / 5 |

## Conclusion

The evaluation demonstrates strong overall performance with a 94.3% cache hit rate and consistent quality across both test cases. Both RFEs received passing scores (8/10 and 10/10), with appropriate decisions made regarding revision needs. The pipeline correctly identified and documented weaknesses in business justification (WHY criterion) and scope bundling where applicable, though it chose not to auto-revise these issues — a defensible decision given the limitations of automated remediation for evidence-based claims.

astefanutti (approved):
Thanks @danmcp!

PR #144, opened 2026-07-27
RHAIFIRST-313: Replace diff binary with difflib in report generation

EmilienM:
## Summary

- Replace `subprocess.run(["diff", ...])` with Python's built-in `difflib.unified_diff()` in `generate_review_pdf.py`
- Eliminates external `diffutils` dependency that is missing in the openshell CI container
- The `autofix-rfe` pipeline's HTML report generation was silently failing because `submit.py` treats the report as optional

## Context

The `autofix-rfe` CI job runs `submit.py --generate-report` after the OpenShell sandbox is destroyed. At that point, the script executes in the outer openshell container, which does not have `diffutils` installed. The `generate_diff()` function was shelling out to the `diff` binary, causing a `FileNotFoundError`.

The standalone `submit-rfe` jobs already work around this by installing `diffutils` via `microdnf`, but using Python's stdlib `difflib` is a cleaner fix that works everywhere.

Jira: https://redhat.atlassian.net/browse/RHAIFIRST-313

## Test plan

- [x] All 51 existing tests pass (`test_generate_run_report.py` + `test_submit.py`)
- [x] Verified `generate_diff()` produces correct unified diff output consumed by `diff_to_html()`
- [ ] Verify HTML report is produced in next `autofix-rfe` pipeline run

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved review PDF generation by producing file differences directly within the application.
  * Increased consistency and reliability across environments by removing reliance on an external system diff utility.

### DIFF
--- a/scripts/generate_review_pdf.py
+++ b/scripts/generate_review_pdf.py
@@ -1,9 +1,9 @@
 #!/usr/bin/env python3
 """Generate an HTML review report from RFE review artifacts."""
 
+import difflib
 import os
 import re
-import subprocess
 import sys
 from collections import Counter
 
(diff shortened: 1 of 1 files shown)


EmilienM:
taken over by autofix: https://github.com/opendatahub-io/rfe-creator/pull/145

PR #147, opened 2026-08-03
Bootstrap agent definitions before the create phase

danmcp:
## Description
`rfe.speedrun` bootstrapped assess-rfe from inside auto-fix's SETUP phase, which copies the `rfe-scorer` definition into `.claude/agents/` moments before ASSESS tries to launch it. The agent registry is rescanned in the background, so the first wave of the run raced it:

    Agent type 'rfe-scorer' not found. Available agents: claude, Explore,
    general-purpose, Plan, statusline-setup

Models that pass `subagent_type` through literally eat that error and have to improvise a fallback; in the 2026-07-30 eval run it cost a wasted agent launch and left the dispatch loop confused enough that 18 of 20 RFEs were never assessed. Runs on the same harness hit it repeatedly — 22 times in one case.

Move bootstrap to Step 0.5, before Phase 1. The create phase then runs for several minutes between installing the definitions and needing them, which is ample rescan time. Bootstrap is idempotent and auto-fix still runs it in SETUP, so a failure here is not fatal — retry once, then continue.

## How Has This Been Tested?
Tested with eval

## Merge criteria:

## Summary by CodeRabbit

* **New Features**
  * Added an automated bootstrap assessment before the pipeline begins.
  * Failed bootstrap attempts are retried once, with processing continuing if the retry fails.

### DIFF
--- a/.claude/skills/rfe.speedrun/SKILL.md
+++ b/.claude/skills/rfe.speedrun/SKILL.md
@@ -32,6 +32,19 @@ Determine pipeline mode:
 
 If no arguments provided, stop with usage instructions.
 
+## Step 0.5: Bootstrap Dependencies
+
+Run bootstrap early so agent definitions (e.g. `rfe-scorer`) are installed
+in `.claude/agents/` before they're needed in Phase 2. The CREATE phase gives
+the background agent rescan time to register them.
+
+```bash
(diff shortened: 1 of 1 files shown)


danmcp:
Fixed in https://github.com/opendatahub-io/rfe-creator/pull/143/

PR #149, opened 2026-08-04
Make review frontmatter parse and repair correctly

danmcp:
## Description

Two ways review frontmatter breaks, both seen in eval runs, both landing on the same code path.

**1. A colon in a hand-written free-text value.** Review agents were writing the frontmatter block themselves:

    needs_attention_reason: Blocked: needs architecture input

`read_frontmatter` called `yaml.safe_load` unguarded, so this surfaced as a raw `ScannerError` traceback. `update_frontmatter` reads before it merges, which meant `frontmatter.py set` — the tool that exists to write correct frontmatter — was itself blocked by the corruption. The only way out was hand-editing YAML, which is what caused it. Two RFEs hit this in a single eval run.

**2. An empty block followed by a horizontal rule.** The delimiter pattern required at least one newline inside the block:

```python
_FRONTMATTER_RE = re.compile(r"^---\s*\n(.*?\n)---\s*\n", re.DOTALL)
```

`(.*?\n)` cannot match nothing, so on an empty block (`---\n---\n`) it consumes the closing delimiter and keeps expanding to the next `---` in the body. When the body is a review with a markdown score table, the table separator becomes the closing delimiter and the table is handed to `yaml.safe_load`:

```
yaml.scanner.ScannerError: while scanning a block scalar
```

This crashed `frontmatter.py set` seven times in the 2026-08-10 initiative-speedrun run, every time a review agent wrote the body before stamping frontmatter.

The two are coupled: the repair path for (1) slices the body on that same regex, so without the regex fix it recovers a truncated body and turns the crash into **silent content loss**. On the captured INIT-006 review it kept 59 of 93 body lines, dropping the heading and the score table. Fixing them together is what makes the repair non-lossy.

Changes:

- `review-agent.md` Step 3 now says to write the body only, with no `---` block. Step 4 creates the frontmatter. `fetch-agent.md` already words it this way.
- `read_frontmatter` raises `ValidationError` with the file, line, and offending text on one line. Callers already handle `ValidationError`, so this replaces a traceback with something readable when agent logs truncate it.
- `update_frontmatter` replaces an unparseable block instead of failing. The body is recovered by regex on the delimiters, so it survives. Validation is the guard: the Step 4 field set alone validates (`auto_revised` defaults to false) so the common case self-heals, while a partial update still fails rather than silently dropping fields. `write_frontmatter` reads the body the same tolerant way, since its data is already validated.
- The closing delimiter is anchored to the start of a line and the block is allowed to be empty, so a `---` inside the body is never mistaken for it. `[ \t]*` rather than `\s*` so a blank line after the delimiter stays in the body.
- An empty block is valid frontmatter with no fields, so it returns `({}, body)` with the delimiters consumed. Returning the full content there would leave `write_frontmatter` stacking a second block on the stale one.

No behavior change for well-formed files.

## How Has This Been Tested?

`uv run pytest tests/` — 774 passed. `ruff check` and `ruff format --check` clean.

`TestReadFrontmatter` and `TestUpdateFrontmatterRepair` cover both triggers: the unquoted-colon block (error message locates the line, repair keeps the body, partial updates still refused, file untouched when refused) and the empty-block-plus-table input (body preserved, a horizontal rule in a normal body left alone, scalar blocks, CRLF, and a rewrite over an empty block leaving one block rather than two).

Also replayed the real failure end to end: the exact `frontmatter.py set` command that crashed seven times in the run now succeeds against the captured INIT-006 review file, with the body byte-identical at 93 lines and a single frontmatter block in the output.

## Merge criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed or empty review metadata, including clearer validation errors and safer recovery when metadata must be replaced.
  * Reviews with invalid metadata are now recorded as errors without interrupting processing.
  * Prevented body separators and Markdown content from being mistaken for metadata delimiters.

* **Documentation**
  * Updated review instructions to generate only the review body before metadata is added separately.

* **Tests**
  * Added coverage for malformed metadata, recovery behavior, empty blocks, line endings, and continued error handling.

### DIFF
--- a/.claude/skills/initiative-review/prompts/review-agent.md
+++ b/.claude/skills/initiative-review/prompts/review-agent.md
@@ -23,9 +23,10 @@ python3 scripts/frontmatter.py schema initiative-review
 ## Step 3: Write Review File
 
 Write `artifacts/initiative-reviews/{ID}-review.md` with the body structure below — just
-the body, no `---` frontmatter block. Step 4 creates the frontmatter. A hand-written
-block breaks the moment a value contains a colon, and `frontmatter.py` cannot then
-read the file to repair it.
+the body, no `---` frontmatter block. Step 4 creates the frontmatter. Writing the
+body only avoids corruption entirely: a hand-written block breaks the moment a
+value contains a colon. (`frontmatter.py set` can recover a corrupted block, but
+don't rely on it.)
 
    ## Assessor Feedback
    <Full rubric feedback verbatim from assessment result>
--- a/.claude/skills/initiative-review/prompts/review-agent.md
+++ b/.claude/skills/initiative-review/prompts/review-agent.md
@@ -22,7 +22,10 @@ python3 scripts/frontmatter.py schema initiative-review
 
 ## Step 3: Write Review File
 
-Write `artifacts/initiative-reviews/{ID}-review.md` with this body structure:
+Write `artifacts/initiative-reviews/{ID}-review.md` with the body structure below — just
+the body, no `---` frontmatter block. Step 4 creates the frontmatter. A hand-written
+block breaks the moment a value contains a colon, and `frontmatter.py` cannot then
+read the file to repair it.
 
    ## Assessor Feedback
    <Full rubric feedback verbatim from assessment result>
--- a/.claude/skills/rfe.review/prompts/review-agent.md
+++ b/.claude/skills/rfe.review/prompts/review-agent.md
@@ -20,9 +20,10 @@ python3 scripts/frontmatter.py schema rfe-review
 ## Step 3: Write Review File
 
 Write `artifacts/rfe-reviews/{ID}-review.md` with the body structure below — just
-the body, no `---` frontmatter block. Step 4 creates the frontmatter. A hand-written
-block breaks the moment a value contains a colon, and `frontmatter.py` cannot then
-read the file to repair it.
+the body, no `---` frontmatter block. Step 4 creates the frontmatter. Writing the
+body only avoids corruption entirely: a hand-written block breaks the moment a
+value contains a colon. (`frontmatter.py set` can recover a corrupted block, but
+don't rely on it.)
 
    ## Assessor Feedback
    <Full rubric feedback verbatim from assessment result>
--- a/.claude/skills/rfe.review/prompts/review-agent.md
+++ b/.claude/skills/rfe.review/prompts/review-agent.md
@@ -19,7 +19,10 @@ python3 scripts/frontmatter.py schema rfe-review
 
 ## Step 3: Write Review File
 
-Write `artifacts/rfe-reviews/{ID}-review.md` with this body structure:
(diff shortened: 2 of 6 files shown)


danmcp:
Eval run with this change:

# Evaluation Summary
**Skill**: initiative-speedrun  
**Model**: claude-opus-4-6  
**Run ID**: 2026-08-11-opus-4-6  
## Overall Result
**PASS**
## Judge Results
| architecture_context_used | check | 1.0 pass_rate | min_pass_rate>=0.85 | ✅ PASS |
| files_exist | check | 1.0 pass_rate | min_pass_rate>=1.0 | ✅ PASS |
| frontmatter_valid | check | 1.0 pass_rate | min_pass_rate>=1.0 | ✅ PASS |
| initiative_quality | llm | 4.3125 mean | min_mean>=3.5 | ✅ PASS |
| pipeline_flow | check | 1.0 pass_rate | min_pass_rate>=1.0 | ✅ PASS |
| recommendation_consistency | check | 1.0 pass_rate | min_pass_rate>=1.0 | ✅ PASS |
| revision_quality | llm | 4.75 mean | min_mean>=3.5 | ✅ PASS |
| run_report_exists | check | 1.0 pass_rate | min_pass_rate>=1.0 | ✅ PASS |
## Execution Metrics
| Metric | Value |
|--------|-------|
| Cases | 16 |
| Duration | 3811.9s (63.5 min) |
| Cost | $36.67 |
| Turns | 635 |
| Cache Hit Rate | 92.0% |
| Cost per Turn | $0.058 |
| Cost per MTok | $1.25 |
| Output Tokens per Turn | 559.0 |
## Per-Case LLM Scores
| Case | initiative_quality | revision_quality |
| case-001-outreach-initiatives | 4 | 5 |
| case-011-add-workflows-api-sdk | 5 | 5 |
| case-012-ambient-capabilities-second-set | 5 | 5 |
| case-013-scaleup-test-matrix-dynamic | 5 | 5 |
## Conclusion
The initiative-speedrun skill running on claude-opus-4-6 passed all configured quality thresholds. All structural judges (files_exist, frontmatter_valid, pipeline_flow, recommendation_consistency, run_report_exists, architecture_context_used) achieved perfect pass rates. The two LLM judges exceeded minimum quality thresholds with initiative_quality scoring 4.31/5 (threshold 3.5) and revision_quality scoring 4.75/5 (threshold 3.5). The pipeline processed 16 cases in 63.5 minutes with a 92% cache hit rate, demonstrating efficient prompt caching. Ten of sixteen cases achieved perfect initiative quality scores (5/5), indicating the pipeline consistently produces high-quality Initiatives.

danmcp:
@astefanutti Your improvements look solid. The coderabbit findings look real but don't seem like blockers. Up to you if you want to fix or defer.

Thanks for taking a look and the additions!

astefanutti:
@coderabbitai review

astefanutti (approved):
Thanks!

PR #155, opened 2026-08-11
Make two more initiative eval cases exercise the revise path

danmcp:
## Description

Recalibrates two existing Initiative eval cases so the pipeline is forced down the revise path to add more coverage. No cases added or removed — the dataset stays at 16.

The two cases below take the opposite ends of that path — one revision that should succeed, one that structurally cannot.

### case-007 — recoverable revision

This case was **already** tagged `prescriptive` with `expected_recommendation: revise`. It scored `open_to_how: 1`, total 9, and passed — the annotation said revise, the pipeline said submit.

The cause is the technology list. Prometheus, DCGM Exporter and Grafana are on the [assess-initiative rubric](https://github.com/opendatahub-io/assess-rfe)'s allowlist of established RHOAI technologies, where naming them counts as platform vocabulary rather than architecture prescription — exactly as the rubric intends. The case was never prescriptive in the way it claimed to be.

Swapped to Telegraf / InfluxDB / Superset / Kapacitor: all off the allowlist, all with obvious alternatives, and framed as a decision that is closed and not to be reopened. Also strengthened WHY (61% of platform cloud spend, sub-20% utilization on the training pool, four capacity escalations) and added explicit in/out scope, so `open_to_how` is the only zero and the case fails purely on the no-zeros rule at a total of ~8.

This one is recoverable. The revise agent's "Reframe, don't remove" rule covers it directly, so a revision should convert the mandated stack into suggestive language and lift the score. That exercises `check_content_preservation.py`, the `-removed-context.yaml` block classification, the re-assess cycle, `before_score` vs `score`, `auto_revised: true`, and the `initiative-auto-revised` label path.

### case-013 — unrecoverable revision

This was the lowest-information case in the set: a 311-character `clarifying_context`, every annotation null, and `infrastructure` as its only tag — already covered by cases 007, 008 and 014.

Kept the domain and rewrote it around a circular justification. It now has a specific deliverable, explicit in/out scope, and an explicitly open implementation, but the entire WHY is that the matrix is static and should not be. Expected `why: 0`, everything else strong.

**Withholding the evidence is not enough on its own.** An earlier draft of this case said only that nobody had counted the blocking, no complaint had been filed, and no numbers existed. It scored `why: 1` in the 2026-08-10 run and again in the 2026-08-11 run, and the assessor's rationale was explicit both times — it credited "four concrete failure modes ... with plausible mechanisms" and read the no-data admission as a well-reasoned engineering argument, which the rubric scores as a 1.

None of those four mechanisms — configuration drift, manual combinatorial management, slow turnaround, no single source of truth — appear anywhere in the input. The create agent derived all four from the deliverable itself. A well-specified WHAT implies its own structural WHY, so staying silent on the evidence just leaves the write-up free to reconstruct the argument from the deliverable.

So the case now **negates** those mechanisms rather than staying silent on them: the list is edited about twice a year, the last hardware retirement was handled in the same PR that decommissioned the nodes, no nightly run has ever failed on a configuration that no longer exists, a matrix change is a one-line edit that lands the same day, and the two consuming teams have never asked for an axis they did not get. The write-up cannot assert drift or turnaround pain without contradicting its source, which leaves bare assertion: the matrix is static and we would rather it were dynamic.

`what`, `scope` and `open_to_how` are untouched, so `why` remains the only candidate zero — and a single zero routes to `revise`, not `reject` ([review-agent.md](.claude/skills/initiative-review/prompts/review-agent.md) reserves reject for 3+ zeros).

This one cannot be fixed by revision — [revise-agent.md](.claude/skills/initiative-review/prompts/revise-agent.md) explicitly forbids inventing evidence — so it should mark the gap, set `needs_attention=true` with a `needs_attention_reason`, and stop at the 2-cycle cap. That is the terminal state the eval has never produced, and it runs the `preserve_review_state.py` save/restore round-trip that the `auto_revised` bug in #154 was hiding in.

Note that case-012 now covers the regression end of that path organically (5 → 4 → `autorevise_reject` in the 2026-08-11 run), so case-013's remaining job is specifically the `needs_attention_reason` path.

### Design constraints

Neither case trips a revise bypass in [filter_for_revision.py](scripts/filter_for_revision.py): `right_sized: 0` routes to split and `feasibility: infeasible` skips revision, so both are avoided.

`expected_feasibility` stays null on both. Off-standard tooling on 007 may legitimately read as `indeterminate`, which does not bypass revision — asserting `feasible` would add a failure mode without testing anything. Following the existing annotation convention, only the targeted criterion is filled in; the rest stay null.

## How Has This Been Tested?

- All 16 cases round-trip through `yaml.safe_load`; annotation keys and `expected_scores` keys match the existing schema exactly; no duplicate prompts across the dataset.
- `scripts/validate_batch_input.py --type initiative --strict` over a batch assembled from all 16 inputs: `ERROR_COUNT=0 WARNING_COUNT=0 VALID=true`.
- `uv run pytest tests/` — 757 passed. No Python touched.
- Whether case-013 actually reaches `why: 0` is only knowable from a live run. The material the assessor credited in the two prior runs has been removed, but this is the third calibration attempt at that criterion.

## Merge criteria:

## Summary by CodeRabbit

* **Evaluation Updates**
  * Refined GPU observability and dynamic scale-up matrix assessment scenarios.
  * Added clearer requirements, scope boundaries, and implementation context for both initiatives.
  * Updated assessment outcomes, recommendations, scoring, and review tags to reflect identified issues and revision needs.
  * Documented concerns including prescriptive technology choices, unsupported evidence, circular justification, and insufficient operational rationale.

### DIFF
--- a/eval/initiative-dataset/cases/case-007-gpu-utilization-observability/annotations.yaml
+++ b/eval/initiative-dataset/cases/case-007-gpu-utilization-observability/annotations.yaml
@@ -1,16 +1,23 @@
 difficulty: null
 expected_alignment: not_assessed
 expected_feasibility: null
-expected_pass: null
+expected_pass: false
 expected_recommendation: revise
 expected_scores:
   what: null
   why: null
   scope: null
-  open_to_how: null
+  open_to_how: 0
   right_sized: null
 expected_total: null
-known_issues: []
+known_issues:
+  - 'Mandates a settled implementation stack (Telegraf, InfluxDB, Superset, Kapacitor) and
+    tells engineering not to reopen it. Should score open_to_how 0 and fail on the no-zeros
+    rule despite strong what/why/scope.'
+  - 'Designed to exercise the recoverable revise path: prescriptive technology choices can be
+    reframed into suggestive language, so revision should lift the score rather than flag for
+    a human.'
 tags:
   - infrastructure
   - prescriptive
+  - revise-candidate
--- a/eval/initiative-dataset/cases/case-007-gpu-utilization-observability/input.yaml
+++ b/eval/initiative-dataset/cases/case-007-gpu-utilization-observability/input.yaml
@@ -1,22 +1,27 @@
 clarifying_context: 'We are spending a lot on GPU clusters and have almost no visibility
-  into how well they are actually being used. People request nodes for KServe inference
-  endpoints or training operator jobs, saturate them for an hour, and then the hardware
-  sits idle until someone else comes along. We do not even have basic utilization
-  dashboards, so nobody can answer simple questions like which vLLM deployments are
-  over-provisioned or whether we could consolidate training workloads onto fewer nodes.
+  into how well they are actually being used. The last three months of billing put GPU
+  nodes at 61 percent of platform cloud spend, and the two spot checks we ran found
+  average utilization under 20 percent on the training pool. People request nodes for
+  KServe inference endpoints or training operator jobs, saturate them for an hour, and
+  then the hardware sits idle until someone else comes along. Four capacity escalations
+  last quarter ended the same way -- we bought more nodes because nobody could show that
+  the ones we had were free.
 
-  The idea is to build an observability layer that collects GPU metrics at the pod and
-  node level, feeds them into our existing monitoring stack, and surfaces dashboards that
-  show utilization trends over time. We also want alerting for sustained low utilization
-  so we can right-size allocations. The implementation should use Prometheus with DCGM
-  Exporter for GPU metric collection, Grafana for the dashboards, and TimescaleDB as
-  the long-term metrics store for historical trend analysis. Eventually we would like
-  to tie this into the scheduling layer so the platform can bin-pack more aggressively,
-  but that is a follow-on -- the first milestone is just getting the data flowing and
-  visible. We think this could save us a meaningful chunk of our cloud bill once teams
-  can actually see what they are using.'
+  What we want to deliver is an observability layer that collects GPU metrics at the pod
+  and node level, retains them long enough for trend analysis, and surfaces dashboards
+  that answer the questions we cannot answer today: which deployments are over-provisioned,
+  which pools sit idle overnight, and where we could consolidate. It also needs alerting
+  on sustained low utilization so teams get told to right-size instead of finding out at
+  the next bill review. In scope: metric collection, retention, the dashboards, and the
+  low-utilization alerts. Out of scope: acting on the data automatically -- feeding it to
+  the scheduler for tighter bin-packing is a follow-on initiative -- and CPU or memory
+  accounting, which is already covered elsewhere. The technical decisions here are made
+  and closed: Telegraf runs as the collection agent on every node, InfluxDB is the metrics
+  store, Superset renders the dashboards, and Kapacitor evaluates the alerting rules.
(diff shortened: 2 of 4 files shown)


danmcp:
Eval run with this change:

# Evaluation Summary
**Skill**: initiative-speedrun  
**Model**: claude-opus-4-6  
**Run ID**: 2026-08-11-opus-4-6  
## Overall Result
**PASS**
## Judge Results
| architecture_context_used | check | 1.0 pass_rate | min_pass_rate>=0.85 | ✅ PASS |
| files_exist | check | 1.0 pass_rate | min_pass_rate>=1.0 | ✅ PASS |
| frontmatter_valid | check | 1.0 pass_rate | min_pass_rate>=1.0 | ✅ PASS |
| initiative_quality | llm | 4.3125 mean | min_mean>=3.5 | ✅ PASS |
| pipeline_flow | check | 1.0 pass_rate | min_pass_rate>=1.0 | ✅ PASS |
| recommendation_consistency | check | 1.0 pass_rate | min_pass_rate>=1.0 | ✅ PASS |
| revision_quality | llm | 4.75 mean | min_mean>=3.5 | ✅ PASS |
| run_report_exists | check | 1.0 pass_rate | min_pass_rate>=1.0 | ✅ PASS |
## Execution Metrics
| Metric | Value |
|--------|-------|
| Cases | 16 |
| Duration | 3811.9s (63.5 min) |
| Cost | $36.67 |
| Turns | 635 |
| Cache Hit Rate | 92.0% |
| Cost per Turn | $0.058 |
| Cost per MTok | $1.25 |
| Output Tokens per Turn | 559.0 |
## Per-Case LLM Scores
| Case | initiative_quality | revision_quality |
| case-001-outreach-initiatives | 4 | 5 |
| case-011-add-workflows-api-sdk | 5 | 5 |
| case-012-ambient-capabilities-second-set | 5 | 5 |
| case-013-scaleup-test-matrix-dynamic | 5 | 5 |
## Conclusion
The initiative-speedrun skill running on claude-opus-4-6 passed all configured quality thresholds. All structural judges (files_exist, frontmatter_valid, pipeline_flow, recommendation_consistency, run_report_exists, architecture_context_used) achieved perfect pass rates. The two LLM judges exceeded minimum quality thresholds with initiative_quality scoring 4.31/5 (threshold 3.5) and revision_quality scoring 4.75/5 (threshold 3.5). The pipeline processed 16 cases in 63.5 minutes with a 92% cache hit rate, demonstrating efficient prompt caching. Ten of sixteen cases achieved perfect initiative quality scores (5/5), indicating the pipeline consistently produces high-quality Initiatives.

astefanutti:
Carried into #193 by cherry-pick (your commit and authorship intact), together with a re-anchoring of the new case-020, so that all three go through one calibration eval; if 007 and 013 fail first review as designed they get the `revision-expected` tag the coverage gate now enforces (added in #190 after your PR). If you would rather land #155 on its own, say so and I will drop the cherry-pick.

danmcp:
> If you would rather land #155 on its own, say so and I will drop the cherry-pick.

No objection at all to your approach. Closing this one.

astefanutti:
The two recalibrated cases landed on main via #193 (your commit cherry-picked with authorship intact, both now tagged `revision-expected` after the calibration run confirmed they fail first review on the targeted criterion). This PR can be closed as superseded whenever you like.

PR #160, opened 2026-08-21
ci-prod: generate the review diff with difflib, not the diff binary

astefanutti:
Targeted backport of the core fix from #145 (RHAIFIRST-313) to `ci-prod`.

## Why the bug is still live in production

#145 replaced `generate_review_pdf.py`'s `subprocess.run(['diff', ...])` with `difflib` — but it merged to `main`, and the autofixer CI clones this repo at `ci-prod` (`CLAUDE_REPO_BRANCH: "ci-prod"`). The openshell image has no `diff` binary, so every production run still hits:

```
FileNotFoundError: [Errno 2] No such file or directory: 'diff'
```

Seen in [job 16022657385](https://gitlab.com/redhat/rhel-ai/agentic-ci/rfe-autofixer/-/jobs/16022657385) (2026-08-21). The job stays green because `submit.py` wraps HTML generation in a warning — the impact is that **no production run publishes its HTML companion report**; the YAML report is unaffected.

## Why a backport rather than a cherry-pick

`ci-prod` is 103 commits behind `main`; a direct cherry-pick of #145's commits conflicts because `main`'s `generate_diff` has since grown hardening (`_safe_artifact_path`, `_open_nofollow`, size caps) entangled with main-only code. This ports only the load-bearing change into ci-prod's existing function shape: read both files, strip frontmatter as before, and produce the unified diff with `difflib.unified_diff` — no temp file, no subprocess, no external binary. The dead `subprocess` import is removed.

Trailing newlines are normalized on both sides so difflib doesn't emit mid-diff `\ No newline` markers — matching what `diff -u` effectively produced.

## Verification

- Functional check against a synthetic original/revised pair: changed case emits a correct unified diff, unchanged case returns `""` (same as `diff -u`).
- ci-prod test suite: 463 passed, 1 failed — the failure (`test_dry_run_skips_split_child_hashes`) pre-exists on unmodified `ci-prod` and is untouched by this change.

### DIFF
--- a/scripts/generate_review_pdf.py
+++ b/scripts/generate_review_pdf.py
@@ -5,7 +5,6 @@ from collections import Counter
 import json
 import os
 import re
-import subprocess
 import sys
 import yaml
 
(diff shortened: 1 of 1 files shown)


astefanutti:
Closing in favour of advancing ci-prod to main wholesale. ci-prod is 103 commits behind and content-wise a strict subset of main (verified: every ci-prod-only commit is content-covered there), and 25+ of the missing PRs are production-relevant and interdependent — so per-fix backports like this one re-author each change and would conflict with the advance itself. The diff-binary fix lands with the advance in its full #145 form (difflib + path/symlink hardening). Until then the impact stays contained: HTML companion reports are skipped with a warning; YAML reports unaffected.

PR #166, opened 2026-08-24
Stop curl reading files, and the input reaching the shell

danmcp:
Ports two fixes found reviewing the same workflow in
opendatahub-io/strat-creator#57. That copy is a PR; this one is already on `main`.

## 1. `curl --form` can be made to upload a file

`--form` treats a value beginning with `@` as a local file to upload.
`variables[GITHUB_HEAD_REF]` carries the PR's branch name, which anyone who can
open a PR controls — and `@/etc/passwd` is a valid git branch name:

```
$ git check-ref-format --branch "@/etc/passwd"
@/etc/passwd
```

Only a bare `@` as the *whole* refname is disallowed, not as a path component. So
the runner can be made to send a local file to GitLab. Demonstrated with a local
listener:

```
--form        "variables[REF]=@victim.txt"  →  filename="victim.txt" + file contents
--form-string "control=@victim.txt"         →  the literal string
```

`--form-string` disables `@`, `<` and `;type=` parsing outright. Every field here
is a literal string, so all eight change, including `token=`.

## 2. The dispatch input reached the shell through `${{ }}`

`${{ inputs.pr_number }}` is expanded before bash runs, so `type: number` does not
constrain it — that is a UI hint, not something the dispatch API enforces. It goes
through `env` now, which is what the *next* step already did.

Worth being clear on severity: `workflow_dispatch` needs write access, so anyone
who could exploit this could also edit the workflow. It is cheap and it makes the
first step consistent with the second.

## Verification

YAML parses, no `${{ }}` remains inside any `run:` block, and the step env
resolves as intended. No behavior change to the trigger itself.

### DIFF
--- a/.github/workflows/rfe-creator-eval.yml
+++ b/.github/workflows/rfe-creator-eval.yml
@@ -47,8 +47,14 @@ jobs:
         id: pr
         env:
           GH_TOKEN: ${{ github.token }}
+          # Through env, not `${{ }}` in the script: a workflow_dispatch input is
+          # expanded into the shell before bash sees it, and `type: number` is a
+          # UI hint rather than a constraint the API enforces. The step below
+          # already reads its values this way.
+          REPO: ${{ github.repository }}
+          PR_NUMBER: ${{ inputs.pr_number }}
         run: |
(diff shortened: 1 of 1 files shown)


astefanutti (approved):
Thanks @danmcp!

PR #171, opened 2026-09-04
Validate pipeline ids before they reach a shell; pin REASSESS_FIXUP's id file

astefanutti:
## Description

Relates to #146 (RHAIFIRST-320), which stays open for rework — this PR carries the part of it that is worth landing now, generalized, and records why one of its changes was not taken.

**1. Validate ids in `_read_ids`.** Every id list `pipeline_state.py` reads is interpolated into `shell=True` commands (`advance()`, `cmd_run_phase`, `pre_script`/`vars` substitution), so a `tmp/*-ids.txt` file is a shell-injection and path-traversal vector. `_read_ids` now validates against a type-neutral grammar `<PREFIX>-<digits>` — every local and Jira id shape the pipeline produces (`RFE-001`, `INIT-004`, `RHAIRFE-2685`, `RHOAIENG-9876`) and nothing else — and aborts with a `[validate-ids]` message naming the file. Differences from #146: its regex was RFE-only (`^(RFE-\d+|RHAIRFE-\d+)$`, which would reject every initiative id on main), and it validated at two call sites, leaving `advance()` uncovered.

**2. Pin `REASSESS_FIXUP`'s `ids_file` — #146's change there is not a fix.** `REASSESS_RESTORE` narrows the reassess set through `filter_for_revision` into `tmp/pipeline-revise-ids.txt`; `REASSESS_REVISE` revises exactly that subset; `REASSESS_FIXUP` must run `check_revised` over the same subset. Switching it to `tmp/pipeline-reassess-ids.txt` would check items nobody revised that cycle. The phase config now says so in a comment, and `TestReassessFixupIds` pins it for both types plus the narrowing behaviour of `advance()`.

**Not carried over:** #146's rewrite of every `PHASE_CONFIG` entry into list form. The same outcome (no shell anywhere) is reached in the second commit by splitting at the two execution sites instead, leaving the string templates for the type registry to regenerate.

Test fixtures that used placeholder ids (`A`, `B`, `X`, `Y`) now use `RHAIRFE-N` so they pass validation; behaviour under test is unchanged.

## How Has This Been Tested?

- `python3 -m pytest tests/ -k "not integration"`: 901 passed (51 new: `TestValidateIds`, `TestReassessFixupIds`, `TestValidateStateValues`, `TestNoShell`, `TestScriptOutputIsValidated`, `TestStateValidatedBeforeDecisionScripts`).
- Integration suite (jira-emulator) run locally; see the CI checks.
- `ruff check` / `ruff format --check` clean.
- Manual: `printf 'RHAIRFE-1\n$(whoami)\n' > tmp/pipeline-active-ids.txt` followed by `python3 scripts/pipeline_state.py next-action` on a FETCH-phase state exits 1 with `[validate-ids] rejected 1 invalid id(s) in tmp/pipeline-active-ids.txt: ['$(whoami)']` and runs nothing.

## Review follow-ups (CodeRabbit, 2026-09-04)

All folded into the two commits above (commit 1: id and state validation; commit 2: shell removal). Kept here as the review trail.

- **ASCII digits** — `\d` is Unicode-aware in Python, so `RHAIRFE-١` passed; the grammar is now `[A-Z][A-Z0-9]*-[0-9]+`, with Arabic-Indic and fullwidth digit tests.
- **Bounded log preview** — the `[validate-ids]` message shows at most five rejected values truncated to 40 characters, so a malformed line is diagnosable without being echoed verbatim.
- **State values formatted into the REPORT command** — `cmd_run_phase` formats `{start_time}` and `{batch_size}` from `tmp/pipeline-state.yaml` and selects the phase table by `type`; `_validate_state_values` now checks all three before use (a tampered value would otherwise become extra argv tokens).
- **`shell=True` removed from both execution sites** (the outside-diff major). No command template uses any shell feature, so `_run_script` and `cmd_run_phase` now split the template with `shlex` and pass argument lists. SETUP's `a & b & wait` becomes a `commands` list launched concurrently with `Popen`; exit codes are logged but not fatal, which is exactly what the old form did (`wait` with no arguments always returns 0), so behaviour is preserved — fail-fast SETUP belongs to the extensibility plan's `validate_types.py --verify` gate. A test asserts no `shell=True` remains in the module. This is done without #146's 573-line `PHASE_CONFIG` rewrite: the string templates stay, so the registry-driven launcher (design-proposals/work-item-types-unified.md §10 PR-2) can still regenerate them.
- **Ids parsed from script output are validated too** (second CodeRabbit review, outside-diff major). `_read_ids` covered ids read from files, but ids parsed from a decision script's stdout did not pass through it: the three `filter_for_revision.py` sites in `advance()` split the output straight into `_write_ids` and `_save_originals` (which builds paths from ids — a CWE-22 traversal), and `_parse_line_ids` fed `REASSESS`/`SPLIT`/`ERRORS` ids to `_write_ids` the same way. Validation now happens at every boundary where ids enter the pipeline: `_ids_from_output`, `_parse_line_ids`, `_write_ids` (backstop) and `_save_originals` itself. Regression tests drive `advance()` at all three sites with malformed output and assert it exits before `_save_originals` runs and before the revise file is written. The dispatch-loop fixtures had space-separated `KEY=` lines while the real scripts emit comma-separated lists (`collect_recommendations.py:72-84`); they now use the real format.
- **State validated before `advance()` and the phase-table lookup** (third CodeRabbit review, outside-diff major). `_validate_state_values` ran only in `cmd_run_phase`; `advance()` formats `state["type"]` into the `--type` flag of the decision scripts and `cmd_next_action` runs `pre_script`/`post_verify` from the phase table, so a modified state file could still inject *arguments* (CWE-88) even with no shell. It now runs at the top of `advance()` and inside `_get_config()`, which every phase-table consumer goes through; tests assert a tampered `type` never reaches `_run_script` from either path.
- **Validator hardening** (fourth CodeRabbit review, three findings on `_validate_state_values` itself): `start_time` is parsed with `strptime` after the shape regex, so `2026-02-30T25:61:61Z` is rejected; a non-string `type` (`type: []`) is a controlled `[validate-state]` rejection instead of a `TypeError` at the `PIPELINE_TYPES` membership test; `batch_size` strings must be ASCII digits (`str.isdigit()` accepts `"²"`, which `int()` rejects in `_wave_size`). Leap days and year boundaries stay accepted.
- **Positive `batch_size`** (fifth CodeRabbit review, minor): the integer branch accepted `0` and negatives and the string branch `"0"`, which `_wave_size` silently turns into one id per wave; `batch_size` must now be a positive integer, with tests for `0`, `-1` and `"0"`.

## Merge criteria:

## Summary by CodeRabbit

* **Bug Fixes**
  * Rejects malformed or unsafe pipeline IDs, state values, and decision-script outputs before processing or file creation.
  * Prevents invalid data from reaching pipeline commands and reassessment workflows.
  * Runs setup steps concurrently and reports individual command failures without interrupting the phase.
  * Improves command handling to prevent unintended shell interpretation.

* **Tests**
  * Expanded coverage for validation, state handling, command execution, setup configuration, and end-to-end pipeline scenarios.

### DIFF
--- a/scripts/pipeline_state.py
+++ b/scripts/pipeline_state.py
@@ -22,6 +22,7 @@ Usage:
 import argparse
 import glob
 import os
+import re
 import shutil
 import subprocess
 import sys
@@ -299,6 +300,11 @@ def _build_phase_config(pipeline_type):
         "REASSESS_FIXUP": {
             "type": "script",
             "command": fixup_cmd,
+            # Deliberately the revise file, not the reassess file: REASSESS_RESTORE
+            # narrows the reassess set through filter_for_revision into
+            # tmp/pipeline-revise-ids.txt, REASSESS_REVISE revises exactly that
+            # subset, and the fixup must check the same subset. Pinned by
+            # tests/test_pipeline_state.py::TestReassessFixupIds.
             "ids_file": "tmp/pipeline-revise-ids.txt",
         },
         # --- Collect + Split ---
@@ -401,7 +407,8 @@ def _build_phase_config(pipeline_type):
 
 
 def _get_config(state):
-    """Get PHASE_CONFIG for the pipeline type in the current state."""
+    """Get PHASE_CONFIG for the pipeline type in the current state (validated first)."""
+    _validate_state_values(state)
     return _build_phase_config(state.get("type", "rfe"))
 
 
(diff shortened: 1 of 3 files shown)


astefanutti:
@coderabbitai review

astefanutti:
@coderabbitai review

astefanutti:
@coderabbitai review

astefanutti:
@coderabbitai review

astefanutti:
@coderabbitai review

astefanutti:
@coderabbitai review

astefanutti:
@coderabbitai review

astefanutti:
@coderabbitai review

PR #178, opened 2026-09-08
Snapshot, bootstrap and fetch adopt the registry (design PR-2c)

astefanutti:
## Summary

PR-2c of the work-item-types design (§10 item 2), **stacked on #177** (which sits on #176): the snapshot and bootstrap configs read the registry, and `fetch_issue --fetch-all` becomes type-aware. **Behavior is byte-identical for every existing invocation.** Three commits.

1. **`snapshot_fetch.SNAPSHOT_CONFIG`** and **`bootstrap_snapshot.BOOTSTRAP_CONFIG`** derive per type from the descriptors (ignore and quarantine labels, snapshot prefix, report prefix, item key), same keys and order; the rfe snapshot prefix that serves as the default keyword argument of the snapshot readers and writer is bound once at import, so signatures are unchanged. The design invariants in `docs/snapshot-incremental-fetch.md` are untouched: nothing changes in what is hashed, persisted, selected or merged, nor in file naming. The bootstrap run-directory probe still reads the **rfe** prefix for every type, as today; making it per-type would change which run directories an initiative bootstrap considers, so it stays a documented, pinned residue for a deliberate follow-up.
2. **`fetch_issue --fetch-all`** gains `--type` (default rfe, additive): directories, id field and the comments companion (written only when `companions.comments` is true) come from the descriptor. `--help` changes only by the new option block. New tests cover both layouts against stubbed Jira responses.
3. **Pins and baseline**: retired pins move to the MIGRATED list, the rfe-only probe keeps an explicit residue pin, prefix baseline 21 → 14 occurrences (11 → 9 files).

## Equivalence evidence

- Golden runs against the pristine PR-2b base: bootstrap_snapshot offline modes on three real results-repo run directories with a deterministic mock Jira; a frozen-time harness over the snapshot load, diff, update and fetch paths on real snapshots; fetch_issue `--fetch-all` with identical stubbed responses; `--help` for all three scripts. Identical stdout, exit codes, produced files and JQL strings (fetch_issue differs only by the documented `--type` block). Reproduced independently by a reviewer and spot-checked by me (config dicts, signatures and help output compared against the base).
- Unit suite 1717 passed; **emulator integration suite 140 passed**; ruff clean; `validate_types` OK; prefix lint within the tightened baseline.

## Deferred on purpose (behavior-changing, need the eval bar rather than golden runs)

Per-type bootstrap probe; `jira_utils.strip_metadata` heading regex rendered from the prefix union; the fast-detection config allowlist; the descriptor-driven bootstrap with `rubric.ref` pinning (§7.3). Each is recorded in the design document.

## Summary by CodeRabbit

- **New Features**
  - Extended “fetch all” workflows to support selecting different work-item types.
  - Generated files, directories, identifiers, snapshot names, and optional comment files now follow each type’s configured format.
  - Command-line type choices and snapshot handling automatically reflect available registered types.

- **Documentation**
  - Updated provider guidance and type documentation with configuration requirements, snapshot settings, and multi-type fetching details.

### DIFF
--- a/docs/type-provider-guide.md
+++ b/docs/type-provider-guide.md
@@ -5,12 +5,15 @@ authoritative text lives. Design: `design-proposals/work-item-types-unified.md` 
 §3.2.1 (binding override), §3.3 (gates), §3.4 (provider obligations), §3.5 (discovery), §3.7
 (cross-type chain).
 
-**Status (PR-2b): 21 scripts read the registry at import** — the "Adoption status" table in
+**Status (PR-2c): 24 scripts read the registry at import** — the "Adoption status" table in
 [`types/README.md`](../types/README.md) lists them and what is still pending. Since PR-2b the
 artifact schemas (`artifact_utils.SCHEMAS`), the scan / rename / parse helpers and the poll phase
 table (`check_review_progress.PHASE_CHECKS`) are projections too, so a registered type gets its
 `<type>-task` / `<type>-review` schemas, its `<poll_prefix><phase>` rows and the generics without
-a code change. Every per-type value a pending script still carries is pinned by test to its
+a code change; since PR-2c so are the snapshot tables (`snapshot_fetch.SNAPSHOT_CONFIG`,
+`bootstrap_snapshot.BOOTSTRAP_CONFIG`) and the `fetch_issue.py --fetch-all` layout (`--type`,
+default `rfe`), so a registered type gets its hard-filter labels, snapshot / run-report file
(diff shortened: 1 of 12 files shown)


astefanutti:
@coderabbitai review

astefanutti:
@coderabbitai review

astefanutti:
@coderabbitai review

astefanutti:
@coderabbitai review

PR #181, opened 2026-09-14
Batch mapping form and per-type parent keys (PR-3b)

astefanutti:
PR-3b of the [PR-3 plan](https://github.com/opendatahub-io/rfe-creator/pull/172) (design §5 rung 2, Q14). Follows #180 (PR-3a). The legacy bare-list batch, which is the form the eval harness and every existing user feed, stays byte-identical on stdout and exit code; the only stream change on a valid input is one stderr line when a type is explicit.

## What lands

- **Both batch root forms.** `validate_batch_input.py` and `next_rfe_id.py --from-batch` accept the legacy bare list and `{type: <t>, items: [...]}`. The type comes from the PR-3a ladder (`type_registry.resolve`): `--type` (default now `None`) → mapping `type:` → legacy default `rfe`, silent. `--type` disagreeing with `type:` is a hard error (D1); a per-item `type` key is a hard error telling the author to split the batch (D2). For the validator these are protocol errors (`ERROR_COUNT=1`, `VALID=false`, exit 1); the allocator exits 2 with the message on stderr and allocates nothing.
- **One parser, one join.** `type_registry.read_batch()` is the single batch parser shared by `resolve()` and both scripts; `Descriptor.parent_key_pattern` is the single anchored join of `conventions.parent_key_patterns`, and `artifact_utils`' task schema delegates to it, so the task schema and the batch rule agree by construction (the permanent schema golden is unchanged).
- **Parent keys and priorities per resolved type.** Initiative batches newly accept `INIT-\d+` parents; the error text is rendered from the descriptor (`'parent_key' must match one of RHAISTRAT-\d+, RHOAIENG-\d+, INIT-\d+`); an rfe entry carrying `parent_key` is an unknown-field warning (rfe declares no `parent_key`), which `--strict`, as the bodies use, still blocks. The priority enum is the resolved type's.
- **Allocator mapping form.** `next_rfe_id.py --from-batch` on a mapping derives prefix and directory from `type:`; explicit `--prefix`/`--dir` that contradict it are a hard error; the list form with the initiative flags is unchanged.
- **D3 for protocol scripts.** Both scripts' stdout is machine-read, so the resolve line goes to **stderr**, and only when the type did not come from the default: `TYPE RESOLVED: initiative (--type)` / `(batch type)`.
- **Speedrun bodies.** One prose paragraph after the unchanged Mode A example in both bodies. The rfe body's validator call now passes `--type rfe` so a mapping declaring another type fails there, before any id is allocated or agent runs (the initiative body already passed `--type initiative`).
- **Lint baseline shrinks:** `validate_batch_input.py` loses its two literal prefixes.

## Visible changes, exhaustively (from the A/B against main)

Valid inputs: stdout and exit code identical for every list-form invocation (20/20 harbor batches with `--strict`, 16/16 initiative one-entry lists, all allocator runs, both `--help`s except the validator's, whose description is its now-accurate docstring). Stderr gains exactly one `TYPE RESOLVED: <t> (--type)` line when `--type` is passed, which both speedrun bodies now do; the legacy default is silent in every environment (malformed `JIRA_PROJECT` / `RFE_CREATOR_BINDING_*` values are not read, pinned). Piped inputs (`/dev/stdin`, process substitution) are read once.

Error paths: D2 per-item `type` was an unknown-field warning, now a hard error; `INIT-` parents accepted for initiatives; rfe malformed `parent_key` was error + warning, now warning only; usage errors of `next_rfe_id.py` exit 2 with one line instead of a traceback; root-shape messages reworded; mapping-form errors are new.

## Proof
`python3 -m pytest -q` → 2258 passed (2131 on main); `validate_types`, prefix lint (baseline shrunk), ruff green. Three-lens review (neutrality A/B of 234 paired runs, adversarial correctness, scope/docs) with 8 findings fixed before opening. **Initiative eval dispatched on this PR** (the parent-key flow runs in every batch); the rfe eval is optional per the plan since the list form is unchanged apart from the one stderr line.

## Not in this PR
Frontmatter `type:`/`tracker_ref:` writers, post-fetch `(project, issue_type)` verification, effective binding in the writers and the run-report `binding:` key (PR-3c). The wait-for-wave escalation (D14) is its own PR. The mapping form does not yet allocate from an overridden `LOCAL_PREFIX` (PR-3c wires the effective binding into the writers).

## Summary by CodeRabbit

* **New Features**
  * Batch files now support both legacy lists and structured `{type, items}` input.
  * Type-aware validation applies the correct fields, priorities, and parent-key rules for each batch type.
  * Batch ID generation derives prefixes and task locations from the resolved type.

* **Bug Fixes**
  * Improved handling of conflicting or invalid type declarations, malformed batch structures, and unsupported fields.
  * Standardized validation and resolution behavior across batch workflows.

* **Documentation**
  * Updated guides and type documentation with supported batch formats and type-resolution behavior.


astefanutti:
@coderabbitai review

PR #182, opened 2026-09-14
Bound the wave barrier: retry or escalate a stalled wave (D14)

astefanutti:
Decision **D14** of the [PR-3 plan](https://github.com/opendatahub-io/rfe-creator/pull/172) and design **R2**: the wave barrier must be bounded before PR-3c and before any new type is registered. This is a deliberate **behaviour PR**, the only one in the PR-3 series: with no stall nothing changes; after a configurable no-progress window a stuck wave is retried or escalated instead of looping until the job timeout.

## The problem
`wait-for-wave` runs the 90-second poll and exits 3 while any `(poll phase, id)` slot is pending; the orchestrator re-runs it. A subagent that dies silently or writes nothing leaves its slot pending forever. This happened repeatedly on weak-model eval runs (scorer misfiles, subagents that never spawned) and the only remedies were a hand-written marker or the 4-hour timeout. An escalation branch (`fix/wave-stall-escalation`, live-validated on two glm-5.3-flash runs in August) is the reference; this PR ports its tracker and reset-on-progress mechanics onto main and replaces its marker writing.

## What lands
- **Progress tracker** (`tmp/pipeline-wave-progress.yaml`): the count of terminal slots and when it last increased. Reset on any progress, re-checked when the 90-second poll returns, and dropped whenever `next-action` or `set-wave` writes a new wave file, so a new wave always starts a fresh window.
- **Policy per pipeline phase** (`WAVE_STALL_POLICY`, pinned complete by test):

| Phases | Policy | Why |
|---|---|---|
| FETCH, ASSESS, REVIEW, REASSESS_ASSESS, REASSESS_REVIEW, SPLIT_ASSESS, SPLIT_REVIEW, SPLIT_REASSESS, SPLIT_RE_REVIEW | **retry, then escalate** | one output file each; a second run overwrites |
| REVISE, REASSESS_REVISE, SPLIT_REVISE, SPLIT | **escalate only**, window ×2 | edits a task in place or mints children; a second agent would double-edit or duplicate |

- **Knobs:** `PIPELINE_WAVE_STALL_SECS` (900; 0 disables the guard entirely), `PIPELINE_WAVE_RETRY_CAP` (2, per phase and id, `tmp/pipeline-stall-retries.yaml`).
- **Retry** leaves the id pending, bumps its counter and exits 0; the orchestrator's `next-action` re-derives a wave from the still-pending ids and launches a fresh agent, which is what usually heals a hung subagent.
- **Escalation is the existing post-barrier contract, never a placeholder.** The registry-derived error-stub review is written through `verify_phase.write_error_stubs` (`error: <phase base>_stalled`, retryable by ERROR_COLLECT); split parents get the non-retryable `split_not_attempted:` class, a failed `-split-status.yaml` in the shape `split_collect` expects, and `needs_attention`; revise records `revise_stalled` on the real review without claiming a revision. The id leaves the wave file and the phase's ids file, so post_verify runs on the survivors and `advance` proceeds. A marker write that fails on a schema-invalid review falls back to replacing the frontmatter with the stub.
- **One stderr line per stall**, e.g. `wait-for-wave: STALL in ASSESS (assess+feasibility): no wave slot reached a terminal state for 900s (window 900s, policy retry); escalating RHAIRFE-1002 (retry cap 2 reached) -> assess_stalled error-stub review, removed from the wave and tmp/pipeline-active-ids.txt`.
- **`AGENTS.md`** (which `CLAUDE.md` links to) gains one sentence in the pipeline constraint so the orchestrator treats a `STALL` line followed by exit 0 as normal and continues with `next-action` rather than improvising.
- **`verify_phase.write_error_stubs`** extracted from `verify()` (same argv, same stdout) and hardened: a merge that fails on a half-written review is replaced rather than swallowed.

## Proof
- `python3 -m pytest -q` → 2190 passed (2131 on main). New: the **bounded barrier test** drives `cmd_wait_for_wave` and `next-action` with a fake clock, a never-completing id and a fake exit-3 poll, and proves retry → cap → escalation, the stub equal field for field to `validate_types.build_error_stub`, removal from wave and ids files, `advance` proceeding, post_verify on survivors only, and `error_collect` queuing the retry. Plus reset-on-progress inside a poll, split and revise escalate-only paths against their real consumers, both shipped types, window 0, fresh tracker per new wave, no trace when nothing stalls, `state.py clean` wiping the files.
- Three-lens review (correctness with an 83-case adversarial harness walking the real orchestrator loop, production safety, scope/docs): 12 findings, all fixed here except two named follow-ups below.
- No-stall neutrality: identical `check_review_progress` argv, exit codes and stdout versus main.

## submit.py guards (added after the CodeRabbit review)
- **Phase 1** skips an Archived parent whose review carries the stall guard's `split_not_attempted: wave stalled …` marker or a `*_stalled` stub, printing `SKIP split-submit - review error …; children were not reviewed, left for an operator`, so children a stalled split agent minted later in the job are never split-submitted unreviewed. submit.py's own not-attempted markers (`Jira preflight failed`, `split phase aborted`) are still attempted on a manual re-run.
- **Phase 2** treats an error-bearing review as having no feasibility verdict: no feasibility label is added or removed, and the item can never auto-approve. **Visible change:** today's `*_failed` stubs stop receiving `rfe-creator-feasibility-pass` for an assessment that never ran (they keep needs-attention).
- **Guard hardening from the same review:** a progress record with the right signature but no usable timestamp is treated as absent (it used to crash every later poll); escalation removes only ids whose marker is on disk, and an id whose marker cannot be written stays in the wave and ids files while `wait-for-wave` exits 1 with an `ESCALATION FAILED … fix the reviews directory and re-run` line, so the failure is loud rather than silent.

## Live validation (fault-injected eval, throwaway #183)
A standard eval cannot exercise this code because nothing stalls on Opus, so a throwaway branch set `PIPELINE_WAVE_STALL_SECS=60` through the eval config's `execution.env` and ran one weak-draft case at full effort. The window reached the pipeline (`stall_window=60s retry_cap=2` printed at init). On the ASSESS wave the scorer finished inside the first poll and the feasibility agent did not; the second poll returned with no new completion, the barrier printed `wait-for-wave: STALL in ASSESS (assess+feasibility): … policy retry; re-dispatching RFE-001 (attempt 1/2)` and exited 0, and the orchestrator continued with `next-action` exactly as the `AGENTS.md` sentence instructs. The re-dispatched wave completed, the weak draft was revised and re-reviewed, and the run passed ($5.45). The no-stall control run (quick mode, two cases) was byte-normal. The escalation path was not reached live, since the original agent always finished; it is covered by the bounded barrier test and the reviewers' harness. Note recorded in the doc: a re-dispatch relaunches every agent of that id's wave, not only the stuck slot.

## Promotion
Behaviour PR: merge → fast-forward `ci-stage` → one `autofix-rfe-stage-dry` (expected: no stall, byte-identical log) → `ci-prod`, then watch the scheduled runs. The guard's first production effect will only be seen on a run that actually stalls.

## Summary by CodeRabbit

- **New Features**
  - Added protection against stalled pipeline waves, automatically retrying eligible work and escalating items that remain stuck.
  - Added configurable stall timeouts and retry limits, with clear outcomes for affected workflows.

- **Bug Fixes**
  - Pipeline execution now continues normally after handled wave retries or escalations.
  - Error-marked reviews no longer receive feasibility approvals, and stalled split parents are skipped safely.
  - Improved fallback handling when existing review records cannot be updated.

- **Documentation**
  - Documented stall detection, retry and escalation behavior, configuration, outcomes, and limitations.


astefanutti:
@coderabbitai review

astefanutti:
@coderabbitai review

astefanutti:
Re the two outside-diff comments in review 5208260687:

- **Normalize persisted retry counts** — done in 9c7065c: `_read_stall_retries` keeps only non-negative non-bool ints; null / string / bool / float / negative values read as 0 and non-string keys are dropped, so a stalled wave can neither fail on `None < cap` nor have a retry bought or denied by a bogus value. Tests: `TestStallRetryCounters`.
- **Fence stale retry attempts** — documented as a limitation in `docs/wave-stall-guard.md` (9c7065c) rather than changed here. The scripts have no handle on a subagent to cancel, and attempt-scoped output paths change the agents' output contract (prompt edits), which is a follow-up. The retry policy already makes the race equivalent by construction for the phases it admits (same agent over the same staged input); the one ordering that can matter is a REVIEW retry whose stale first-pass review lands after the item was revised and re-reviewed, which needs an agent a full window late that then finishes minutes later still, and is visible after the fact through the run report's provenance fields.
- **Path traversal** (inline) — done in d469dd5.

PR #183, opened 2026-09-15
[throwaway] D14 stall guard fault-injection eval — do not merge

astefanutti:
Throwaway on top of #182: sets PIPELINE_WAVE_STALL_SECS=120 in eval.yaml's runner env so a quick-mode RFE eval trips the wave stall guard on legitimate agents. Purpose: read the transcript for STALL lines, re-dispatch, escalation and the orchestrator's reaction. Will be closed after the run; never merge.


astefanutti:
Served its purpose: run 2 (execution.env window 60 s, one weak-draft case) tripped the guard live on the ASSESS wave — retry path re-dispatched, orchestrator continued with next-action, pipeline passed. Branch kept for a possible escalation-path run.

PR #187, opened 2026-09-15
Effective binding in the writers, is_existing from tracker_ref, run-report binding key (PR-3c-iii)

astefanutti:
Last of the three PR-3c PRs ([plan](https://github.com/opendatahub-io/rfe-creator/pull/172); design §3.2.1 rules c, d, e, g; §5 `is_existing`; decisions D11, D12). **Stacked on #186** (base `feat/fetch-verification`); it retargets to `main` when #186 merges. Because the lint workflow and CodeRabbit only trigger for PRs targeting `main`, this PR shows no CI or CodeRabbit review until that retarget happens; the full gate (pytest incl. the emulator suites, validate_types, the prefix lint, ruff) is green locally on the head commit, and both evals run against the head regardless of the base. With no override set, production is byte-identical: same dry-run plan, same Jira calls, same files; only the run report gains an additive header key.

## What lands
- **Resolution in the writers.** `submit.py`, `split_submit.py` and `check_conflicts.py` resolve the type (`--type` default now `None`), print the D3 line only for an explicit `--type` (a split child spawned by submit stays silent), prove ownership with `assert_registered_binding`, and refuse a `JIRA_PROJECT` / `JIRA_ISSUE_TYPE` shorthand binding with one line pointing at the typed variables (the shorthand stays a `resolve` CLI convenience: the artifact layer builds its grammars at import and cannot honour it, and a live submit under it would create an issue it then cannot rename).
- **Effective binding.** Project, issue type and write prefix come from `Descriptor.binding(env)`: the dry-run sentinel is `<PROJECT>-DRY`, split children are created in the effective project, `_TRACKER` is keyed by type name (D12). The task id grammar and the rename guard use the effective key prefixes (overridden write prefix first, descriptor prefixes kept as read prefixes); the parent-key grammar of the task schema and the batch validator gains the overridden write prefix. No override → the schemas golden and every emulator golden are unchanged.
- **`is_existing := tracker_ref`.** A task is existing when its frontmatter `tracker_ref` is owned by the resolved type; a `tracker_ref` owned by another type is a hard error (never an update, never a create); a pre-migration artifact without the field falls back to effective key-prefix membership. The remote key every write targets is `tracker_ref` when present, in `submit.py` and `check_conflicts.py` alike.
- **Pre-write witness checks.** Before updating an issue, the writer's own fetch requests `project` and `issuetype` and verifies them against the pairs the key may witness: the effective pair always, and the descriptor pair for a key carrying a descriptor read prefix, so items created before an override stay updatable. `split_submit` applies the same check to the parent before it comments, labels, links or closes it.
- **Run report** gains an additive `binding:` header key `{tracker, project, issue_type, source}` (D11); `bootstrap_snapshot` tolerates it.
- **Hermetic tests:** the conftest strips `RFE_CREATOR_BINDING_*` and the shorthand variables so the suite does not depend on the developer's shell.

## Fail-closed rules added after the CodeRabbit review
An existing issue that cannot be verified is never written. `submit.py` skips such an item (a failed conflict check or witness fetch) with the reason in the plan and leaves it unprocessed for the next run, where it used to warn and update anyway; `split_submit.py` refuses to write when the parent verification raises, classified per parent or systemic by `_classify_exit`; `check_conflicts.py` exits 2 on a Jira fetch failure instead of reporting no conflicts (its documented code for API failures). **Visible change on a transient Jira error:** one skipped item and a retry next run instead of an unverified update.

## Deliberately deferred
`LOCAL_PREFIX` is registry-only until the artifact grammar reads `binding()['local_id_pattern']` (D6); the how-to says so. Snapshot header metadata (D11, second half) stays a follow-up that updates the snapshot invariants doc first.

## Proof
`python3 -m pytest -q` → 2748 passed (2583 on #186), including a new overridden-project emulator suite (`RFE_CREATOR_BINDING_RFE_PROJECT=KONFLUX`: fetch, dry-run plan with `KONFLUX-DRY`, live create and rename in KONFLUX, split children in KONFLUX, `check_conflicts`, run report `binding: {project: KONFLUX, source: env}`; refusals: the initiative pair under an rfe override, a foreign `tracker_ref`, the shorthand, an Epic parent, a malformed variable as one line). Three-lens review (production A/B against #186 on live emulator seeds, the override scenario end to end, scope/docs) → 15 findings; the workflow's fixer stalled, so they were triaged by hand: all applied except wording-only items that were folded into the docs. **Both evals PASS on this PR**: RFE `rfe_quality` 4.40 / `revision_quality` 4.72 / coverage 0.96 over 25 cases (re-run after #188 fixed the speedrun body's missing clarifying-context forwarding, which had produced a spurious coverage 0.0 on the first run and is unrelated to this PR); initiative `initiative_quality` 4.31 / `revision_quality` 4.75 / coverage 1.0 over 16 cases. After merge: stage dry run, ci-prod, one watched prod run.

## Summary by CodeRabbit

- **New Features**
  - Support project, issue-type, and key-prefix overrides through typed binding settings.
  - Reports now record the effective tracker binding used for each run.
  - Split operations and artifact handling follow the selected effective binding.

- **Bug Fixes**
  - Added validation to prevent updates, fetches, or splits when tracker bindings do not match.
  - Improved handling of tracker references and overridden project keys.
  - Invalid shorthand configuration is rejected before tracker operations begin.

- **Documentation**
  - Expanded guidance for effective bindings, deployment overrides, and directing types to another project.


astefanutti:
@coderabbitai review

PR #193, opened 2026-09-18
Initiative weak drafts: re-anchor case-020, carry in the two #155 recalibrations for calibration

astefanutti:
Follow-up to #190 on the initiative weak-draft cases, two things in one calibration run:

1. **case-020 re-anchored.** The #192 eval scored it WHAT, WHY and Scope at 0, which the review contract routes to `reject`, so the tagged case never entered the revise path (the coverage miss in that run). The input now describes a concrete weekly digest with the same circular-WHY anchor, so WHY is the only targeted zero, consistent with the two-zero rule documented after the case-019 review.
2. **#155 carried in.** Dan's commit recalibrating `case-007` (off-allowlist prescriptive stack, closed decisions) and `case-013` (negated WHY) is cherry-picked with his authorship. His write-up is the sharper form of the lesson behind #190 — withholding evidence lets the create step reconstruct WHY from a well-specified WHAT, so the input has to negate the mechanisms — and the two cases were never merged. They keep his `revise-candidate` tags in this PR.

**Plan for the calibration run on this PR:** if 007 and 013 fail first review as designed, a follow-up commit tags them `revision-expected` (six enforced weak drafts of 20; at 0.92 one may still go unrevised) and updates the fragment's calibration note; if either passes, it stays `revise-candidate` and the reason goes in the README. Either way case-020 is expected back on the revise path.

Supersedes #155 once merged (noted there).

## Summary by CodeRabbit

* **Documentation**
  * Updated weak-draft evaluation guidance and calibration status for initiative-speedrun cases.
  * Clarified the expected outcomes, scoring criteria, known issues, and revision paths for several evaluation scenarios.

* **Tests**
  * Refined GPU observability, dynamic test-matrix, and cluster health digest scenarios with more specific scope, evidence, constraints, and expected review outcomes.
  * Added or updated tags identifying cases that should fail initial review or require revision.


astefanutti:
Calibration run (`2026-09-18-opus-4-6`, job 16588308415): every weak draft failed first review on the criterion it targets and was revised — 007 HOW 0 (8→10), 013 WHY 0 (8→9), 017 WHY 0 (8→9), 018 HOW 0 (8→10), 019 HOW 0, 020 WHY 0 (8→9) — `revision_coverage` 1.0. d96ac21 therefore tags 007 and 013 `revision-expected` (six of 20; five must be revised at 0.92) and updates the calibration note, README and arithmetic test.

The run's single breach, `revision_flag_consistency` on case-019, is the AISDLC-33 residual, not this dataset: the item was restored correctly at COLLECT and its cycle-2 review agent wrote the review twice more afterwards, minutes later. Fixed by moving the final reconcile to REPORT in the follow-up PR.

astefanutti:
@coderabbitai review

